PR #21, opened 2020-10-09
Readme explaining implementation and design

### LINKED ISSUE #20: Readme: explain internals, implementation so that nextstrain team can easily develop docs
Right now the readme just explains the submodule set up. We need to explain the bigger picture of how it all fits together, so that very soon when we are moving into production with these docs, anyone on the nextstrain team can easily get the info they need to know to jump in and contribute changes and improvements. It would also behoove us to include in our documentation of this project some how-tos as we come across handy tricks and methods for accomplishing specific things with sphinx, restructured text, and read the docs - without duplicating information on https://www.sphinx-doc.org/en/master/ and https://docs.readthedocs.io/en/stable/.

eharkins:
Should close #20 for now and allow team members to more easily jump in and provide content, edits, and feedback.


jameshadfield (approved):
Thanks @eharkins -- this is a really good summary of the current situation.

After the merge of this PR we are well positioned to solicit feedback from the group -- there may be big improvements which can be made (as you refer to with the "best of both worlds" section here), but if not then the repo only requires some minor tweaks to be ready for release.

PR #28, opened 2020-10-16
Improve organisation of the installation docs

jameshadfield:
Here we improve the clarity of the installation docs by shifting the
introductory content to the main (index) "Installing Nextstrain"
page. This then naturally leads to the installation pages for various
parts of the nextstrain ecosystem.

The rendered titles of the pages in the sidebar are improved.

The content in the "local installation" page is reduced to remove
duplication between that and the augur/auspice install pages, which
will help avoid divergence going forward.

PR #30, opened 2020-10-16
Improve sharing section organisation

jameshadfield:
This PR is similar in spirit to #28, but goes slightly further

* The visualisation & sharing section felt too broad, and is here separated into "sharing" and "communicating"
* The fetch documentation recently added to nextstrain.org/docs is added here (see nextstrain/nextstrain.org@04b7da8)
* Duplicated pages describing community URL sharing were pruned
* Instead of the page summarising the different sharing options being one of the pages underneath the "share" section, this lifts it to be displayed when selecting the "share" section.


eharkins (approved):
This all makes sense to me @jameshadfield. I added a commit 3499b92375e613ffd8429867d51b99527a520306 which removes the sharing overview page from where it was moved from, so that it only shows up once at the top level of that section, since I was assuming this was your intention with:
> Instead of the page summarising the different sharing options being one of the pages underneath the "share" section, this lifts it to be displayed when selecting the "share" section.

If not, we can add it back in!

Issue #31, opened 2020-10-19
Improve contribute section


eharkins:
1. titles need distinct names see #22 
2. need pages for all nextstrain tools if we have them for any / if its a comprehensive listing of how to contrib to different pieces
3. https://docs.nextstrain.org/en/migrate/guides/contribute/documentation.html needs rewriting for these docs - some of this writing has already been done in readme so for now maybe we can use that and try to improve it more once our implementation development has settled a bit.
4. We should link to and/or include https://github.com/nextstrain/.github/blob/master/CONTRIBUTING.md

Please add any other improvements needed, I'll try to jump on these ones now!

PR #32, opened 2020-10-20
Improve contribute section

eharkins:
See #31 for the goal of this and see commit messages for implementation details.

My implementation of 
> https://docs.nextstrain.org/en/migrate/guides/contribute/documentation.html needs rewriting for these docs - some of this writing has already been done in readme so for now maybe we can use that and try to improve it more once our implementation development has settled a bit.

uses symbolic links as per https://github.com/sphinx-doc/sphinx/issues/2840#issuecomment-284863143 to include the readme in the build of these docs as a guide for contributing documentation. There are other suggestions for how to do something like this in that thread.


eharkins:
Hm, seems like the symbolic link approach did not translate from my local environment to read the docs' build: https://readthedocs.org/projects/nextstrain/builds/12140142/. I'll look into this.

eharkins:
I tried an alternative method of sourcing the readme in the build here: https://github.com/nextstrain/docs.nextstrain.org/compare/improve-contribute...improve-contribute-readme?expand=1

which was basically this:
1. rename `src` to `docs`
2. move the readme to `docs` dir. Note that this doesn't affect the readme as it shows up here on github. See  https://docs.github.com/en/free-pro-team@latest/github/creating-cloning-and-archiving-repositories/about-readmes:
> If you put your README file in your repository's root, docs, or hidden .github directory, GitHub will recognize and automatically surface your README to repository visitors.

3. now I can include the readme in a toctree without sphinx complaining and without using symlinks. 

However, if you look at https://docs.nextstrain.org/en/improve-contribute-readme/guides/contribute/index.html, the [Contributing to Documentation](https://github.com/nextstrain/docs.nextstrain.org/compare/improve-contribute...improve-contribute-readme?expand=1#diff-3db59702858b06b752965c2b471cb9286184951349b8b714ea0f3bb7849f8955R15) entry (aka the readme) does not show up. I'm not sure why this is, as this approach also worked locally, as did the symlink approach. I couldn't find anything obvious in build logs.

Maybe there are simpler ways or taking a step back maybe including this repo's readme in the sphinx build is not something we want or need at the end of the day. @tsibley do you have experience with this?

One alternative is if we really thing most of the content in the readme belongs in the docs, we can move it to an actual document there and then link to it from the readme.

eharkins:
6973b7a adds a "Contributing to Documentation" page with a link out to the readme: https://docs.nextstrain.org/en/improve-contribute/guides/contribute/documentation.html

This solves the above issue about including the readme, and we can improve on this later as we see fit. @jameshadfield merge away if this works for you!

jameshadfield (approved):
This is a nice improvement. Let's merge this before `migrate` gets merged into `master` 👍

Issue #27, opened 2020-10-15
Shared table of contents across all nextstrain read the docs projects


eharkins:
We achieve #3 using git submodules; see https://github.com/nextstrain/docs.nextstrain.org/tree/migrate#submodules

However, this solution is not ideal; see https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1602020878007600

So, we also use Read The Docs **subprojects**; see https://github.com/nextstrain/docs.nextstrain.org/tree/migrate#how-we-are-currently-using-subprojects 

This issue is a placeholder for a "best of both worlds" approach, which is currently being pursued by @trs: 
>  A subprojects + shared/global table of contents approach is preferred to the current subprojects + submodules + separate table of contents approach.
I'll carve a few days out to finish implementing this, and at the same time, James and Eli will continue to push forward with their remaining tasks. We'll see how timing of the completion of both coincides and make a decision to either launch docs.nextstrain.org with the current approach or use the new approach depending on if the new approach is ready.
If the launch happens with the current approach, then we can switch approaches later, just with a bit more tedium.

For more discussion on that, see https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1602626783056100?thread_ts=1602622975.053700&cid=C7SDVPBLZ

eharkins:
In #36 we move from a submodule approach to a fetching approach where we make a request for a page from it's github URL. Importantly we still hope to replace that approach with the approach mentioned here: a shared table of contents across nextstrain read the docs projects.

tsibley:
Some notes to self on prior art I recently stumbled across that's new to me:

- http://www.virtualroadside.com/blog/index.php/2017/01/14/integrated-subproject-sites-on-readthedocs-org/
- https://github.com/DriesSchaumont/sphinx-subprojecttoctree/issues/65

victorlin:
I'm out of the loop here and have a couple questions:

1. Given that all the projects have pretty different TOCs, what would a shared TOC look like?
2. Is this the only solution to stop fetching docs from other repositories?

tsibley:
The original idea was that you could browse/read the subprojects without losing access to the umbrella project's sidebar/search. For example, if you click on "Nextstrain CLI" from the umbrella project:

![image](https://github.com/nextstrain/docs.nextstrain.org/assets/79913/3eda7455-1b7c-4263-bae6-005e13390f54)

[...]

Issue #14, opened 2020-09-25
Fix links after migrating to read the docs


eharkins:
There are many links which will not work or not make sense after migrating everything to be hosted on the same domain.

Here are some examples:

- Relative paths in documents which are genuinely migrated to this repository (as opposed to those which remain in their own repository and are imported from a submodule), e.g. links to `/docs/*` in nextstrain.org documents.
- "Links out" from one docs domain to another (e.g. https://nextstrain.org/docs/getting-started/local-installation#install-augur-with-python) will no longer make sense; we'll want to change these to internal read the docs paths.
- Relative paths to non-markdown, github hosted files from github hosted docs domains (e.g. on this page https://nextstrain.github.io/ncov/customizing-analysis) will need to be replaced with the full github url so that they work on read the docs.

eharkins:
Hopefully redirects will cover most of this, but there might be some links in nextstrain repo readmes that we will want to change, or at least the language around the links may change, e.g. https://github.com/nextstrain/.github/blob/master/CONTRIBUTING.md#contribute-documentation

genehack:
As `make linkcheck` now runs without error on `docs.nextstrain.org:master`, I am closing this issue.

Issue #3, opened 2020-09-15
"import" documents from other repositories


eharkins:
In some cases, where it makes more sense to store the document in a different repository, we will need to include the document directly from it's home repository; the current plan for this is using submodules. See #2 for an example of this.

eharkins:
We achieve this using git submodules; see https://github.com/nextstrain/docs.nextstrain.org/tree/migrate#submodules

However, this solution is not ideal; see https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1602020878007600

So, we also use Read The Docs **subprojects**; see https://github.com/nextstrain/docs.nextstrain.org/tree/migrate#how-we-are-currently-using-subprojects 

A "best of both worlds" approach is being pursued in the meantime by @trs: see #27

PR #36, opened 2020-11-07
Fetch docs instead of using submodules

eharkins:
As discussed on Slack, we now fetch documents that live in other repos instead of using submodules (at least until #27 happens).

### Related issue(s)  
#3 #14 

### Testing
Build works locally and on rtd, and I addressed all warnings related to relative links etc.
Still would be good to have another pair(s) of eyes to check for broken links etc, and ask the important questions about how this sets us up for success or failure when it comes to #27.


eharkins:
Note that this also involves some changes to other repositories where the relevant documents are hosted; see:

- no documents are hosted in nextstrain.org but I need to check https://github.com/nextstrain/nextstrain.org/pull/226 to see if the redirects there need updating according to this PR
- https://github.com/nextstrain/augur/pull/620
- https://github.com/nextstrain/cli/pull/102
- https://github.com/nextstrain/auspice/pull/1220
- we are still using master branch from ncov repo.

eharkins:
Looks like we will need to update some redirect url entries in https://github.com/nextstrain/nextstrain.org/pull/226/files.

I'll wait to do so until we confirm they are what we want here.

jameshadfield (approved):

PR #39, opened 2020-12-18
Charon v2 design brainstorm

kairstenfay:
This is a collaborative document meant to be used in designing the charon v2 API. 
Note that some work for the charon v2 API has already been started on a development branch at https://github.com/nextstrain/nextstrain.org/pull/235. At the time of opening this PR, that branch exists only as a draft PR and is not ready for merging.

### Related issue(s)  
See the documentation for related issues.

### Testing
The documentation compiles locally, although I'm not sure it lives in the right place.
It might be better to exist under the Auspice documentation, for example.

### DIFF
--- a/src/reference/api/charon.md
+++ b/src/reference/api/charon.md
@@ -1,0 +1,103 @@
+# Charon API
+
+## Unversioned API
+A dynamic server, also unofficially called "v1".
+Previously [documented here](https://docs.nextstrain.org/projects/auspice/en/latest/server/api.html).
+
+
+## V2
+Version 2 (v2) of the charon API will be static instead of dynamic.
+Some of the content on this page comes from the linked discussions and notes below:
+* Tom's notes from [Dec 26, 2018](https://github.com/tsibley/blab-standup/blob/master/2018-12-26.md#proposed)
+* GitHub discussion on [Jan 1, 2019](https://github.com/nextstrain/auspice/issues/687)
+* Tom's notes from [Apr 9, 2020](https://github.com/tsibley/blab-standup/blob/master/2020-04-09.md)
+
+
+### Motivation
+Creating an API in REST format will better enable us to extend the customers of the API which is especially important with respect to how we want to use Nextstrain Groups for outreach and to enable data accessibility.
+For example, we could perform dataset operations via the Nextstrain CLI (see [relevant issue](https://github.com/nextstrain/cli/issues/90)).
+We could eventually build a custom UI for Nextstrain.org to allow dataset management within the browser.
+
+
+### Proposed design
+This new design will maintain complete, backwards compatibility with the unversioned charon API.
+
+action  | method  | v1 (`/charon`)    | v2 (`/charon/v2`)   | notes
+---     | ---     | ---               | ---                 | ---
+list all datasets at top-level | GET | `/getAvailable` | `/dataset` | v1 also lists narratives
+list all datasets at path | GET | `/getAvailable?prefix={datasetPath}` | `/dataset/{datasetPath}` | see note above
+retrieve dataset at path | GET | `/getDataset?prefix={datasetPath}` | `/dataset/{datasetPath}.json` |
+retrieve dataset at path | GET | `/getDataset?prefix={datasetPath}&type={type}` | `/dataset/{datasetPath}/{type}.json` |
+list all narratives at top-level | GET | not implemented | `/narrative`   | for v1, `/getAvailable` returns all narratives
+list all narratives at path | GET | not implemented | `/narrative/{narrativePath}` | see note above
+retrieve narrative at path | GET | `/getNarrative?prefix={narrativePath}&type=md` | `/narrative/{narrativePath}.md` | `.markdown` extension will also be accepted.
+retrieve source info at path | GET | `/getSourceInfo?prefix={datasetPath}` | TBD |
+upload dataset(s) at path | PUT | not implemented | `/dataset/{datasetPath}` | must be specific to dataset
+upload narrative(s) at path | PUT | not implemented | `/narrative/{narrativePath}` | must be specific to dataset
+remove dataset(s) at path | DELETE | not implemented | `dataset/{datasetPath}` | must be specific to dataset
+remove narrative(s) at path | DELETE | not implemented | `narrative/{narrativePath}` | must be specific to narrative
+
+
+#### TODO
+How to implement:
+* redirects (e.g. `/charon/v2/dataset/flu` → `/charon/v2/dataset/flu/seasonal/h3n2/ha/2y`)
+* support old JSON formats (v1, meta+tree)
+* 204 responses (which are possible from `/charon/getDataset` and are relied on by auspice)
+* is our synatx for getting a dataset `/charon/v2/dataset/{datasetPath}/main.json` or is it `/charon/v2/dataset/{datasetPath}.json`?
+
+
+#### New Features
+Some ideas for new features not covered in the table above include:
+
+* ability to filter datasets or narratives (e.g. based on maintainer name)
+
+
+#### Examples
+Complete examples of v2 API requests and their expected response codes will live in the [Nextstrain.org smoke-tests](https://github.com/nextstrain/nextstrain.org/blob/master/test/smoke-test/auspice_client_requests.test.js).
+Some v2 examples are also written in the section here on [Proposed CLI Syntax](#proposed-cli-syntax).
+
+description | v1 request    | v2 request
+---         | ---           | ---
+list all datasets and narratives in community dataset | `/charon/getAvailable?prefix=/community/blab/zika-colombia` | [`/charon/v2/dataset/community/blab/zika-colombia`, `/charon/v2/narrative/community/blab/zika-colombia`] |
+retrieve core dataset | `/charon/getDataset?prefix=/flu/seasonal/h3n2/ha/2y` | `/charon/v2/dataset/flu/seasonal/h3n2/ha/2y`
+retrieve core narrative | `/charon/getNarrative?prefix=/narratives/ncov/sit-rep/2020-05-15&type=md` | `/charon/v2/narrative/ncov/sit-rep/2020-05-15.md`
+
+
+### Proposed CLI Syntax
+The Nextstrain CLI will be a prominent client of the v2 API.
+Considering the desired UX of the CLI can help inform how v2 should be designed.
+The following CLI examples are adapted from [Tom's notes](https://github.com/tsibley/blab-standup/blob/master/2020-04-09.md).
+
+`nextstrain remote ls nextstrain.org`
+
+    → GET nextstrain.org/charon/v2/dataset
+
+`nextstrain remote ls nextstrain.org/groups/blab`
+
+    → GET nextstrain.org/charon/v2/groups/blab
+
+`nextstrain remote download nextstrain.org/flu/seasonal/h3n2/ha/2y`
+
+    → GET nextstrain.org/charon/v2/dataset/flu/seasonal/h3n2/ha/2y/main.json
+    → GET nextstrain.org/charon/v2/dataset/flu/seasonal/h3n2/ha/2y/tip-frequencies.json
+    → GET nextstrain.org/charon/v2/dataset/flu/seasonal/h3n2/ha/2y/….json
+
+`nextstrain remote download nextstrain.org/flu`
+
+    → fails; not specific to a dataset; could be supported in future
+
+`nextstrain remote upload nextstrain.org/ncov/ global.json europe.json …`
+
+    → PUT nextstrain.org/charon/v2/dataset/ncov/global.json
+    → PUT nextstrain.org/charon/v2/dataset/ncov/europe.json
+    …
+
+`nextstrain remote rm nextstrain.org/flu/seasonal/h3n2/ha/2y`
+
+    → DELETE nextstrain.org/charon/v2/dataset/flu/seasonal/h3n2/ha/2y/main.json
+    → DELETE nextstrain.org/charon/v2/dataset/flu/seasonal/h3n2/ha/2y/tip-frequencies.json
+    → DELETE nextstrain.org/charon/v2/dataset/flu/seasonal/h3n2/ha/2y/….json
+
+`nextstrain remote rm nextstrain.org/flu`
+
+    → fails; not specific to a dataset; matches existing behaviour
--- a/src/reference/api/index.rst
+++ b/src/reference/api/index.rst
@@ -1,0 +1,13 @@
+
+======================================
+Nextstrain.org API
+======================================
+
+Reference material regarding the Nextstrain.org "charon" API.
+
+.. toctree::
+   :maxdepth: 2
+   :titlesonly:
+   :caption: Table of contents
+
+   charon
--- a/src/reference/index.rst
+++ b/src/reference/index.rst
@@ -11,6 +11,7 @@ Reference guides and API documentation that give implementation and usage detail
    :titlesonly:
    :caption: Table of contents
 
+   api/index
    formats/index
    Augur: A bioinformatics toolkit for phylogenetic analysis <https://docs.nextstrain.org/projects/augur/en/stable/index.html>
    Auspice: An Open-source Interactive Tool for Visualising Phylogenomic Data <https://docs.nextstrain.org/projects/auspice/en/stable/>


tsibley:
Thanks for getting this doc started, @kairstenfay! Probably unsurprisingly, I have a lot of thoughts here, old and new that are stuck in my head and not yet transcribed to paper. It may take a little time to solidify them, but I'll write them all down and share them as soon as I can.

jameshadfield:
Thanks for this @kairstenfay  — it’s really nice to see this taking shape! I spent some time thinking through this and have the following questions. In general, I came at this from two perspectives:

a. I imagined how I would update auspice to use this new API, which I think we’ll want to to do sometime in the future.

b. During development/implementation/maintenance of the current API, there have been lots of little pain points, things that felt overly hard to do and things that felt awkward or hacky to accomplish. I tried to collect these here so that the new design can accommodate them.

Thanks for this work!

---

### 1. General list and retrieve of dataset(s)

Context: currently, when auspice requests resources related to a URL `X` it makes a request for the dataset (`/getDataset?prefix=X`) as well as the available datasets and narratives (`/getAvailable?prefix=X`). In the situation where the dataset is valid (i.e. exists), then the latter request is used to populate the dataset dropdowns in auspice’s sidebar. 

1.1. How would this work for the v2 API in the case where the dataset did exist? From the proposal, it doesn’t seem possible to list datasets in this case, as `/dataset/X` will return the dataset and not the listing of available datasets.

1.2. Importantly, the behavior of the `getAvailable?prefix=X` request is dependent on the prefix (`X`), as this is mapped to a `Source` and the list of returned datasets & narratives are specific to that source.  If we imagine that `X/Y/Z` doesn’t exist as a dataset, then `/charon/dataset/X/Y/Z` will list datasets. Which ones will it list? Will it be similar to path expansion, such that it lists all datasets beneath the hierarchy of `X/Y/Z`, or will it be similar to v1 where it’ll map `X/Y/Z` to a `Source` and list all datasets in that source?

1.3. Currently a `getDataset` API call to a non-existent dataset may either 404 or return a 302 redirect.  As concrete examples, `/charon/getDataset?prefix=/zika/bad` 404s while `/charon/getDataset?prefix=/flu` is a 302 redirecting to `/charon/getDataset?prefix=/flu/seasonal/h3n2/ha/2y`.  On the auspice client side, this is how nextstrain.org/zika/bad shows a 404 splash page and nextstrain.org/flu ends up at nextstrain.org/flu/seasonal/h3n2/ha/2y, respectively. In the current proposal `GET /charon/v2/dataset/flu` would return a listing of datasets. How should auspice navigate itself from /flu to /flu/seasonal/h3n2/ha/2y in this situation?

1.4 In the proposed design table, some (v2) requests include a file suffix and some don’t. It would be nice to have consistency here. 


### 2. Static site support

A major advantage of the v2 API is that “Endpoints can be provided … by a set of files on a static host (e.g. GitHub Pages, S3)”. Could you detail how this would work here, especially as it pertains to directory structure (on the static host) and file suffixes:

2.1 Given a file `myDataset.json` what is the GET call to retrieve this?

2.2 Where does this file (`myDataset.json`) live? Does it have to be in a directory structure `/charon/v2/dataset/`?

2.3 If I create a static file listing the available datasets, where should this be stored (and named) and what is the corresponding GET call to access it? Similarly for a file listing available narratives.


### 3. Authn and Authz

Currently (v1), GET  requests for authenticated users include a cookie (e.g. `nextstrain.org: s:RrjLQU<redacted>`). The nextstrain.org server uses this to authorize requests for private data.

It would be great to make authn a component of the API proposal from the start (and clearly this is part of it since this design follows CLI login work, and PUT / DELETE actions will need auth).

3.1 Will charon v2 calls use the same approach to auth as charon v1, namely the `nextstrain.org` cookie?

3.2 Is the intended API for login/logout documented anywhere? Looking at https://github.com/nextstrain/cli/pull/92/files it appears that the proposed method for the CLI to do this is to talk to cognito directly, but I would have expected that this is to be handled by a (charon?) API call to nextstrain.org.

3.3. As a concrete use case for a nextstrain project in current development, it would be great for [auctor](github.com/nextstrain/auctor) to be able to login, list available (private) narratives for a particular source, access authenticated narratives/datasets and update them all via this API. Would this be possible?


### 4. Versioned requests for resources

Currently the way we handle different versions (or formats) of resources is inconsistent. As an example, we currently have two dataset versions (the “unified” dataset, JSON and a “meta+tree” JSON pair). When a request for `getDataset?prefix=X` is made, only a unified JSON is returned. If the underlying resource is actually a “meta+json” paid, then this is converted server-side to the “unified” format. This means that an older client which only uses the “meta+tree” JSON format couldn’t use the v1 API to access this resource. 

Will there be a way to specify the desired version of the requested resource via this API?


### 5. API surface

Currently the `/charon` API calls originate solely (?) from the nextstrain.org-built auspice client. But we also have some other API calls, namely around the login/logout flow. Is the remit of charon to be dataset/narrative related, or is this flexible? 

Specifically, were we to have a in-browser augur-runner, would it interact with the cloud through this API or a completely separate one?

### 6. Status codes

During discussion of your work changing the status codes of v1 API responses (https://github.com/nextstrain/nextstrain.org/pull/228) it was mentioned that this work would pave the way for the v2 design.

6.1 Are these important to the proposal here?

6.2 Are `204` responses (which are possible from `/charon/getDataset` and are relied on by auspice) part of v2? If not, is there a description of the logic a client such as Auspice would use to replicate current behavior. 


### 7. RESTful versioning of the API

The current proposal uses a slug (`/v2/`) to define the version, which implies future versions would change the slug to `/v3/` etc. As the motivation is to “[Create] an API in REST format”  why not use [content negotiation](https://blog.ploeh.dk/2015/06/22/rest-implies-content-negotiation/) instead? This is similar in spirit to how we version dataset JSONs - via a property within the file rather than a filename slug.


### 8.  get source info

This is TBD in the current spec. For context, currently within nextstrain.org this is only used (only useful) for community and groups sources. The client which calls this API is the nextstrain.org-customised build of auspice — it’s not in Auspice itself, which is why it’s not part of the v1 documentation in auspice’s docs.

### 9. CLI syntax

Thanks for the examples of CLI syntax — these were helpful! It strikes me that the CLI has commands that are similar to charon v1:
* `nextstrain remote ls` <-> `charon/getAvailable`
* `nextstrain remote download` <-> `charon/getDataset`
Whereas the v2 API specifically removes the distinction between these two API calls. Related to point 1 (above), what happens in the following situations?

9 .1 `nextstrain remote ls nextstrain.org/flu/seasonal/h3n2/ha/2y`

9.2. `nextstrain remote download "nextstrain.org/flu/*"` (will this be allowed?)

9.3 In the CLI syntax section, the API requests for the main dataset file is `GET /charon/v2/dataset/{datasetPath}/main.json` which differs from the proposed design table, where the GET request would be `GET /charon/v2/dataset/{datasetPath}.json`. Which one is correct here?

kairstenfay:
Thanks for this thorough feedback, @jameshadfield. I'm going to try to address all your concerns, but I'll post feedback for each section as I go through this list. 

> ### 1. General list and retrieve of dataset(s)
> 
> Context: currently, when auspice requests resources related to a URL `X` it makes a request for the dataset (`/getDataset?prefix=X`) as well as the available datasets and narratives (`/getAvailable?prefix=X`). In the situation where the dataset is valid (i.e. exists), then the latter request is used to populate the dataset dropdowns in auspice’s sidebar.
> 
> 1.1. How would this work for the v2 API in the case where the dataset did exist? From the proposal, it doesn’t seem possible to list datasets in this case, as `/dataset/X` will return the dataset and not the listing of available datasets.
> 
> 1.2. Importantly, the behavior of the `getAvailable?prefix=X` request is dependent on the prefix (`X`), as this is mapped to a `Source` and the list of returned datasets & narratives are specific to that source. If we imagine that `X/Y/Z` doesn’t exist as a dataset, then `/charon/dataset/X/Y/Z` will list datasets. Which ones will it list? Will it be similar to path expansion, such that it lists all datasets beneath the hierarchy of `X/Y/Z`, or will it be similar to v1 where it’ll map `X/Y/Z` to a `Source` and list all datasets in that source?
> 
> 1.3. Currently a `getDataset` API call to a non-existent dataset may either 404 or return a 302 redirect. As concrete examples, `/charon/getDataset?prefix=/zika/bad` 404s while `/charon/getDataset?prefix=/flu` is a 302 redirecting to `/charon/getDataset?prefix=/flu/seasonal/h3n2/ha/2y`. On the auspice client side, this is how nextstrain.org/zika/bad shows a 404 splash page and nextstrain.org/flu ends up at nextstrain.org/flu/seasonal/h3n2/ha/2y, respectively. In the current proposal `GET /charon/v2/dataset/flu` would return a listing of datasets. How should auspice navigate itself from /flu to /flu/seasonal/h3n2/ha/2y in this situation?
> 
> 1.4 In the proposed design table, some (v2) requests include a file suffix and some don’t. It would be nice to have consistency here.

I hope I've understood your feedback here. Let me address 1.4 first. Following your example with prefix `X`, a GET request to `/charon/v2/dataset/X` would list all datasets available at that path. Adding the suffix, e.g. `/charon/v2/dataset/X.json`, retrieves the specified dataset at that path. Therefore, the `.json` suffix is the major distinction there. 

I believe clarifying this also addresses 1.1 and 1.2. 

Re: 1.3. Great call out. The redirecting (302 response) of `/charon/v2/dataset/flu` → `/charon/v2/dataset/flu/seasonal/h3n2/ha/2y` does not work out of the box with the change to the V2 API. I'm not sure off hand how auspice/charon v1 handle redirect requests, but I believe the best approach is to maintain a configuration file of the redirects we want to perform, and have the new API read from there if the original, requested URL (e.g. `charon/v2/dataset/flu`) is mapped in the file.

Please let me know if I misunderstood any of your points in section 1.

kairstenfay:
> ### 2. Static site support
> 
> A major advantage of the v2 API is that “Endpoints can be provided … by a set of files on a static host (e.g. GitHub Pages, S3)”. Could you detail how this would work here, especially as it pertains to directory structure (on the static host) and file suffixes:
> 
> 2.1 Given a file `myDataset.json` what is the GET call to retrieve this?

It depends on where `myDataset.json` lives. Is that a core dataset? Is it part of a public or private Nextstrain group? Is it from a community build? etc. The answer changes depending on the source of the file. 

> 2.2 Where does this file (`myDataset.json`) live? Does it have to be in a directory structure `/charon/v2/dataset/`?

Good question. I believe the goal is to store most of these files on S3, except for those sourced from public GitHub repositories (my understanding on these GitHub sources is pretty fuzzy). 
In effect, we shouldn't see existing files move anywhere, because the behind-the-scenes mapping of "prefixes" (or query parameters) to Sources will remain the same between charon v1 and v2. 

`/charon/v2/dataset` is an arbitrary endpoint that the client must call, and it does not correspond to any directory structure. The endpoint could have just as easily been named something concise like `get` or verbose (and nonsensical) like `/myApi/42/1/1/1/fetch`, etc., and the only thing that would change is how the client must call the endpoint.

> 2.3 If I create a static file listing the available datasets, where should this be stored (and named) and what is the corresponding GET call to access it? Similarly for a file listing available narratives.

Is this how auspice currently behaves? e.g. the `getAvailable` endpoint actually accesses a static file that lists the datasets and narratives? If not, then I may need a minimal working example to understand what you're asking here.

kairstenfay:
> ### 3. Authn and Authz
> 
> Currently (v1), GET requests for authenticated users include a cookie (e.g. `nextstrain.org: s:RrjLQU<redacted>`). The nextstrain.org server uses this to authorize requests for private data.
> 
> It would be great to make authn a component of the API proposal from the start (and clearly this is part of it since this design follows CLI login work, and PUT / DELETE actions will need auth).
> 
> 3.1 Will charon v2 calls use the same approach to auth as charon v1, namely the `nextstrain.org` cookie?

I'm not sure. Perhaps someone with more experience here can chime in? cc @tsibley 

> 3.2 Is the intended API for login/logout documented anywhere? Looking at https://github.com/nextstrain/cli/pull/92/files it appears that the proposed method for the CLI to do this is to talk to cognito directly, but I would have expected that this is to be handled by a (charon?) API call to nextstrain.org.

I was under the assumption that there would be no charon v2 API call for logging in and logging out. Rather, the client would configure their environment with the correct AWS or Nextstrain.org credentials for authn. With the Nextstrain CLI's new `login` feature (https://github.com/nextstrain/cli/pull/92), once a user signs in, their Nextstrain.org Groups credentials are stored on their machine for future access.

> 3.3. As a concrete use case for a nextstrain project in current development, it would be great for [auctor](github.com/nextstrain/auctor) to be able to login, list available (private) narratives for a particular source, access authenticated narratives/datasets and update them all via this API. Would this be possible?

Thanks for writing that use case. This should be possible. Currently the concept of "logging in" only exists on Nextstrain.org and the CLI (https://github.com/nextstrain/cli/pull/92), but those actions should be executable so long as the `auctor` service account or user has the proper credentials configured.

kairstenfay:
> ### 4. Versioned requests for resources
> 
> Currently the way we handle different versions (or formats) of resources is inconsistent. As an example, we currently have two dataset versions (the “unified” dataset, JSON and a “meta+tree” JSON pair). When a request for `getDataset?prefix=X` is made, only a unified JSON is returned. If the underlying resource is actually a “meta+json” paid, then this is converted server-side to the “unified” format. This means that an older client which only uses the “meta+tree” JSON format couldn’t use the v1 API to access this resource.
> 
> Will there be a way to specify the desired version of the requested resource via this API?

Interesting edge case. We should be able to return the"meta+tree" JSON pair on request by tweaking the API call. Perhaps we could add an additional endpoint, something like: `charon/v2/dataset/v1/X`. The mixing in of v2 and v1 is likely to be confusing given that they address two different versions: the former being the charon API version, and the latter being the JSON format version. I'm open to ideas here, but I think we could fairly easily support old JSON formats.

kairstenfay:
> ### 5. API surface
> 
> Currently the `/charon` API calls originate solely (?) from the nextstrain.org-built auspice client. But we also have some other API calls, namely around the login/logout flow. Is the remit of charon to be dataset/narrative related, or is this flexible?

The original design intentions for charon v2 were to be dataset and narrative oriented, but I encourage @tsibley to opine, since he wrote the original design notes. 

Are you asking about extending the functionality of charon v2 beyond datasets & narratives in order to perform a login/logout? or are there other reasons for asking?

> Specifically, were we to have a in-browser augur-runner, would it interact with the cloud through this API or a completely separate one?

I believe an in-browser augur-runner would interact with the new, charon v2 API.

kairstenfay:
> ### 6. Status codes
> 
> During discussion of your work changing the status codes of v1 API responses ([nextstrain/nextstrain.org#228](https://github.com/nextstrain/nextstrain.org/pull/228)) it was mentioned that this work would pave the way for the v2 design.
> 
> 6.1 Are these important to the proposal here?

Sort of. Returning a more appropriate response code (e.g. 404 instead of 500) will enable clients of the charon v2 API (e.g. Nextstrain.org or the Nextstrain CLI) to respond more precisely to failed requests. They can be considered an enhancement of charon v1.

> 6.2 Are `204` responses (which are possible from `/charon/getDataset` and are relied on by auspice) part of v2? If not, is there a description of the logic a client such as Auspice would use to replicate current behavior.

Great question. I don't think anyone has put any thought into that yet. Perhaps we could treat 204s the same way we eventually decide to treat 302s. That is to say, perhaps those custom responses can be mapped in a separate config (JSON?) file.

kairstenfay:
> ### 7. RESTful versioning of the API
> 
> The current proposal uses a slug (`/v2/`) to define the version, which implies future versions would change the slug to `/v3/` etc. As the motivation is to “[Create] an API in REST format” why not use [content negotiation](https://blog.ploeh.dk/2015/06/22/rest-implies-content-negotiation/) instead? This is similar in spirit to how we version dataset JSONs - via a property within the file rather than a filename slug.

I just finished reading through the content negotiation article you linked. I had not heard of that approach before. It appears to me that the client has to specify the version in either approach --either by modifying URLs or modifying the request header. Is there a reason you are inclined to go with content negotiation over URL versioning? I don't have a strong opinion either way, I'm just not able to see a clear advantage of content negotiation for our case. 

One thing I wanted to flag that might be a deal-breaker:
> The most common criticisms against using Content Negotiation for versioning is that it makes the API less approachable. The argument goes that with version information in the URL, you can still test and explore the API using your browser. Once you add the requirement that HTTP headers should be used, you can no longer test and explore the API with your browser. 

I seem to recall that various Nextstrain team members call `getDataset` requests pretty frequently in the browser. We would apparently no longer be able to do with the content negotiation approach to versioning.

kairstenfay:
> ### 8. get source info
> 
> This is TBD in the current spec. For context, currently within nextstrain.org this is only used (only useful) for community and groups sources. The client which calls this API is the nextstrain.org-customised build of auspice — it’s not in Auspice itself, which is why it’s not part of the v1 documentation in auspice’s docs.

Thanks for this context. Do you think get source info should be supported in v2?

kairstenfay:
> ### 9. CLI syntax
> 
> Thanks for the examples of CLI syntax — these were helpful! It strikes me that the CLI has commands that are similar to charon v1:
> 
>     * `nextstrain remote ls` <-> `charon/getAvailable`
> 
>     * `nextstrain remote download` <-> `charon/getDataset`
>       Whereas the v2 API specifically removes the distinction between these two API calls. Related to point 1 (above), what happens in the following situations?
> 
> 
> 9 .1 `nextstrain remote ls nextstrain.org/flu/seasonal/h3n2/ha/2y`

This CLI command would make this v2 API call:
```
→ GET nextstrain.org/charon/v2/flu/seasonal/h3n2/ha/2y
```
It would list all datasets available in the core flu dataset at that path. It would be similar in concept to `getAvailable?prefix=/flu/seasonal/h3n2/ha/2y` except only list datasets and filter them to only datasets beginning with `flu/seasonal/h3n2/na/2y` (for some reason, the equivalent `getAvailable` request does not do this filtering).


> 9.2. `nextstrain remote download "nextstrain.org/flu/*"` (will this be allowed?)

Great question! We could possibly add wildcards as a feature for v2. However, I'm not sure that in your example, `remote download nextstrain.org/flu/*` would do anything that `remote download nextstrain.org/flu` doesn't. 

The details of this haven't really been worked out since it's a CLI design problem, but my guess is it would go something like this.

CLI realizes that the requested dataset to `download` did not have a file suffix, therefore it may match multiple files or directories. So, it first makes this charon v2 request:
```
→ GET nextstrain.org/charon/v2/flu/
```
to retrieve the list of available datasets listed at that path and saves it locally. Then, either via iteration or bulk export, it fetches the files listed under `flu/` using charon v2 API calls like:
```
→ GET nextstrain.org/charon/v2/flu/avian/h5n1/ha/main.json
→ GET nextstrain.org/charon/v2/flu/avian/h5n1/ha/tip-frequencies.json
→ GET nextstrain.org/charon/v2/flu/avian/h7n9/ha/main.json
→ GET nextstrain.org/charon/v2/flu/avian/h7n9/ha/tip-frequencies.json
…
```

> 9.3 In the CLI syntax section, the API requests for the main dataset file is `GET /charon/v2/dataset/{datasetPath}/main.json` which differs from the proposed design table, where the GET request would be `GET /charon/v2/dataset/{datasetPath}.json`. Which one is correct here?

Thanks for that catch. I am not very familiar with the JSON naming schemes. In writing this design doc, I had not considered the `{datasetPath}/main.json` naming scheme. Between that and `{datasetPath}.json`, which seems better to you?

jameshadfield:
Thanks for taking the time to address these @kairstenfay. I think there’s still some design work to be done around use cases that can’t be currently supported but that we want to support, as well as ensuring there’s a path for all current functionality to be supported by the new API. 

### Listing available datasets vs getting datasets, and static site support

> Therefore, the .json suffix is the major distinction there [between getting available and getting the dataset]
> 
> 2.2 Where does this file (myDataset.json) live? Does it have to be in a directory structure /charon/v2/dataset/?
> 
> 2.3 If I create a static file listing the available datasets, where should this be stored (and named) and what is the corresponding GET call to access it? Similarly for a file listing available narratives.

I think there’s some confusion about static sites here. When the idea of charon v2 was first introduced it was focused on auspice (CLI didn’t yet exist). The big improvement over the v1 API was that it would support non-nextstrain.org auspice deploys as a static site, without the need to run a server to dynamically interpret client requests. In other words, the getDataset & getAvailable requests from the auspice client would only need a server which returned static content (e.g. we could run an auspice instance on GitHub pages — see [here](https://blab.github.io/ZEBOV/) for a proof-of-principle of this).

In this case, the static site would store (e.g.) JSON(s) representing the dataset(s) and a JSON (or JSONs?) representing the list of available datasets (etc). I don’t see how it would be possible, with your design, for the static site to get the “available datasets” file here.  Specifically, “the .json suffix is the major distinction” seems to imply that a dynamic server is necessary.

### Auth

> I’m not sure [if charon v2 calls use the same approach to auth as charon v1]

> I was under the assumption that there would be no charon v2 API call for logging in and logging out.

> those actions should be executable so long as the auctor service account or user has the proper credentials configured.

> Are you asking about extending the functionality of charon v2 beyond datasets & narratives in order to perform a login/logout? 

One of the transformative aspects of the v2 API will be to allow privileged access for various parts of nextstrain. Examples include auctor, web access to managing datasets, running builds etc. I’d like this to be front and center in the API design. For instance, how credentials are included in API calls, and how we imagine these are created. Specifically on the latter point, we currently use AWS/cognito, but this may not always be the case; thus the more services that manage their own login/logout flow by talking to cognito directly means the more things we’d have to change if we ever switched to another auth provider.

This relates to point 5 (the API surface) and 8 (sourceInfo) — in other words, is this API by itself going to allow us to support things like auctor & web-access to managing resources, or will they also need other technologies (e.g. direct calls to AWS/cognito)?

### What does “available” datasets mean?

> It would list all datasets available in the core flu dataset at that path. It would be similar in concept to getAvailable?prefix=/flu/seasonal/h3n2/ha/2y except only list datasets and filter them to only datasets beginning with flu/seasonal/h3n2/na/2y (for some reason, the equivalent getAvailable request does not do this filtering).

The current behaviour of `getAvailable?prefix=X` is subtle but really important. The nextstrain.org server does not interpret `X` as a traditional filepath hierarchy. Consider the response of `https://nextstrain.org/charon/getAvailable?prefix=/flu/seasonal/h3n2` — it doesn’t just list all the available h3n2-related datasets, rather it lists all of the core datasets! Specifically, the server maps `X` to a `Source` (in the sense we use that word for in our server-side code) and then returns all available datasets for that Source. Similarly `https://nextstrain.org/charon/getAvailable?prefix=/groups/blab/ncov/tutorial` returns all datasets in that group.


### Other points

> The redirecting (302 response) of `/charon/v2/dataset/flu` → `/charon/v2/dataset/flu/seasonal/h3n2/ha/2y` does not work out of the box with the change to the V2 API. 

This is quite a major part of nextstrain/auspice at the moment. We can definitely change auspice’s behaviour here (see also the next comment, re: `204`) but I think we should have a rough idea of how we’d change it.

Also note that it’s not enough to dynamically interpret the request for (e.g.) `flu` server-side and return the data for `flu/seasonal...`, as the redirects are used by the auspice client to change its URL (e.g. load nextstrain.org/flu and see what happens to the URL!). Note also that the server can’t simply intercept the initial (HTML) request for `/flu`, as this isn’t made if you change datasets via the dropdown. 

> 6.2 Are 204 responses … part of v2?
> Great question. I don’t think anyone has put any thought into that yet… perhaps those custom responses can be mapped in a separate config (JSON?) file.

I think this should be thought about before we proceed. What the server uses internally (e.g. JSON configs) is pretty easy to change down the road, but the HTTP response here should be well defined as it is required for the auspice client to work.

> I am not very familiar with the JSON naming schemes. In writing this design doc, I had not considered the `{datasetPath}/main.json` naming scheme. Between that and `{datasetPath}.json`, which seems better to you?

All (auspice v2) datasets use the latter format.

PR #42, opened 2021-01-09
Guide: how to add a build to nextstrain.org/sars-cov-2

eharkins:
So that there are instructions out there for adding builds to this page on nextstrain.org. 

https://github.com/nextstrain/nextstrain.org/pull/248 is important to note here as it adjusts the formatting of that YAML file a bit. 

Would be appreciated to have folks who have recently added builds give this document a quick read through at https://docs.nextstrain.org/en/adding-a-build/guides/share/nextstrain-registry.html and suggest any edits here. cc @cassiawag @barneypotter24

### DIFF
--- a/src/guides/share/index.rst
+++ b/src/guides/share/index.rst
@@ -11,6 +11,7 @@ Sharing Analyses through Nextstrain
    nextstrain-groups
    community-builds
    fetch-via-urls
+   nextstrain-registry
    Download data <download-data>
 
 
--- a/src/guides/share/index.rst
+++ b/src/guides/share/index.rst
@@ -11,7 +11,7 @@ Sharing Analyses through Nextstrain
    nextstrain-groups
    community-builds
    fetch-via-urls
-   nextstrain-registry
+   sars-cov-2
    Download data <download-data>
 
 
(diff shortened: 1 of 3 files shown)


emmahodcroft:
This a great idea @eharkins ! Thank you for doing this!

The only thing I'd maybe suggest is to add a quick line on what to do if you're adding a build on a state or country where one already exists - that you should adjust the lat-long values for both builds slightly, so the dots don't overlap. We'll need to keep an eye out for this ourselves as we merge PR's too, I suppose!

barneypotter24:
Thanks for making this @eharkins. 

Definitely a very useful addition to the documentation!

I would maybe add to the "hierarchies" section that this refers to the dropdown menus, or possibly just add a labeled picture showing what each YAML element refers to. 

I'm not sure if the YAMLs require a particular tab style (`\t` vs spaces) but that may be good to clarify if it potentially causes issues.

jameshadfield (changes requested):
Thanks @eharkins - this will really help with allowing people to add their builds! Things to add / modify from my point of view:

* Specify that spaces should be used (tabs may work, but we currently use spaces both here and across most/all nextstrain YAMLs & JSONs). 
* Now that https://github.com/nextstrain/nextstrain.org/pull/248's gone in, update text to make it clear that "heirachy entries" are all specified before "build entries"
* Add comment about overlapping GPS coords. May want to mention the script added by the above PR as well?
* (When this PR goes in) update the YAML itself to add a link to this docs page 😄 

Thanks!

eharkins:
Thanks everyone for your feedback! I added some changes which hopefully address most of what's been discussed here, let me know if there's anything else you think I should include.

jameshadfield (approved):

PR #45, opened 2021-02-11
Document how to use the augur index command

huddlej:
Adds a section to the Zika tutorial describing how to run the augur index command and why you would want to do so.


eharkins (approved):
Looks good!

PR #47, opened 2021-02-22
Improve docs to update a local installation

jameshadfield:
Previous instructions had a few shortcomings, including using `pip` rather than our preferred `python3 -m pip`, failed to update `conda` itself, and failed to update the external dependencies of `augur` which can't be installed by `pip` (`mafft` etc)

### DIFF
--- a/src/guides/install/local-installation.md
+++ b/src/guides/install/local-installation.md
@@ -52,10 +52,21 @@ conda deactivate
 ```
 
 
-## Updating `augur`, `auspice`, and `nextstrain`:
+## Updating
+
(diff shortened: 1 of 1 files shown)


huddlej (approved):
Thank you for pushing for this, @jameshadfield. This was a big gap in the docs. I've pushed some revisions to the commands and text to align the update content with the installation content.

We're not quite at the point of using the Augur Bioconda package in the Nextstrain conda YAML, so I reverted the Augur update command to the PyPI version. There's no reason we can't update the conda YAML to use the Bioconda package though and it's on my list of things to do soon.

PR #51, opened 2021-03-02
Update instructions to update Nextstrain with Bioconda

huddlej:
Given that the Nextstrain conda environment will [soon install Augur and the Nextstrain CLI from Bioconda](https://github.com/nextstrain/conda/pull/11), we can simplify the instructions to update the Nextstrain environment by removing references to pip upgrades of Augur and the CLI.

### DIFF
--- a/src/guides/install/local-installation.md
+++ b/src/guides/install/local-installation.md
@@ -72,7 +72,4 @@ conda update --all
 
 # Update Auspice.
 npm update --global auspice
-
-# Pull the latest Docker image for the CLI.
-nextstrain update
 ```
--- a/src/guides/install/local-installation.md
+++ b/src/guides/install/local-installation.md
@@ -67,16 +67,12 @@ Then, update each individual program, as necessary.
 ```sh
 conda activate nextstrain
 
-# Update Nextstrain dependencies (mafft, etc.).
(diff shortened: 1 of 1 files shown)


src/guides/install/local-installation.md
```diff
@@ -67,16 +67,12 @@ Then, update each individual program, as necessary.
...
 
-# Update Augur.
-python3 -m pip install --upgrade nextstrain-augur
-
 # Update Auspice.
 npm update --global auspice
 
-# Update the Nextstrain CLI and pull the latest Docker image.
-python3 -m pip install --upgrade nextstrain-cli
+# Pull the latest Docker image for the CLI.
```
jameshadfield, line 76: One shouldn't need to upate the docker image for a "local" install, but I wonder how hard it would be to have `nextstrain update` also update itself and (conda-installed) augur/auspice?

huddlej: That's a great catch! I didn't notice that inconsistency on the last round of edits. We could move this to the CLI install docs which don't mention the `nextstrain update` command.

I like the idea of having `nextstrin update` update the Docker or conda environments, as needed. @tsibley mentioned something like this would be possible in the existing CLI framework.

tsibley: Yep! The middle-distance plan is to have `nextstrain update` be able to both update itself and a Conda-based environment.

Issue #49, opened 2021-02-25
nextstrain quickstart guide - several issues and need of updates


davidcroll:
Hi,

I'm David and quite new to the team. As a first task, @rneher directed me to review the [Quickstart guide](https://github.com/nextstrain/docs.nextstrain.org/blob/master/src/tutorials/quickstart.md) which is a bit outdated.

I'm checking whether this Quickstart guide is working, and which problems are cropping up. Expect this issue to evolve over the next days, as I'm re-testing a few things.

But... anyway: Feel free to make comments or propose solutions.

__Testing environment__ 

I'm using Xubuntu 20.10 in VirtualBox. The installation is unchanged, save for the `fish` console (which I love) and `augur` from the Ubuntu repository. `augur` from the repo is a problem, as it is quite old. I will be re-testing the whole affair with the more recent `augur` installed through `python3-pip` and from source (`git`).

Here, the various issues:

## 1 - Python version

Revise the guide to mention Python 3.8. Python 3.6 has been released in December 2016. Reminding people to use Python > 3.5 is a bit over-cautious nowadays.

## 2 - PATH issue after python3-pip install of nextstrain-cli

The guide asks to do `python3 -m pip install nextstrain-cli`.

When the user - as told to - enters `nextstrain version` next, then the program isn't found. This was confusing (even to me), and this is a transient error because nextstrain is not installed on PATH.

Solution: Tell the user to log out - and in again.

## 3 - nextstrain.cli version number

When the quickstart guide mentions version 1.4.1 (instead of 3-ish), then it really shows how much development work has been done!

## 4 - Docker installation

The guide mentions: _On Linux, your package manager should include a Docker package_

Well, what am I to do?

I google "install docker on ubuntu", I find a solution, and then...

```bash
sudo apt install docker.io
sudo systemctl enable --now docker

## The recommended test...
sudo docker run hello-world
```

These three lines could be added to the quickstart guide - if they are the _right_ ones. 

## 5 - Checking nextstrain setup

So, we have `nextstrain check-setup`. In contrast to the old guide, nowadays a lot of other conditions are being tested. In my case, I got this result:

```bash

vm@vm ~> nextstrain check-setup
nextstrain-cli is up to date!
 
Testing your setup…
[...]

PR #52, opened 2021-03-06
Consolidate installation docs into a single page

huddlej:
Adds a new top-level page with the installation instructions that provide the single clearest path (and an alternate) for most users. The recommended path uses Conda to bootstrap the user's Python environment and install all Nextstrain tools. Additional (optional) instructions to use the Docker image with the Nextstrain CLI follow this initial installation.

Tasks for this PR:

- [x] Draft new installation guide with single clearest path for most users
- [x] Move content from previous installation docs into more appropriate supporting documentation
- [x] Update the quickstart to link to this new installation guide
- [x] Update links elsewhere in the docs to refer to this new installation guide
- [x] Add [redirects in RTD for previous installation docs](https://readthedocs.org/dashboard/nextstrain/redirects/)
- [x] Consider using UI tabs for Docker OS sections (as in [this example](https://docs.readthedocs.io/en/stable/config-file/v2.html))
- [x] User test these new docs for Conda and Docker use cases


jameshadfield:
@davidcroll would you be able to take a look at the content added by this PR? It was in part motivated by #49, and we're aiming to replace the current installation instructions in the quickstart with a link to the install docs added here.

huddlej:
@davidcroll, I just want to echo @jameshadfield's interest in your feedback on the proposed installation guide. You can [explore this version of the documentation](https://nextstrain--52.org.readthedocs.build/en/52/install-nextstrain.html) and if you comment on this PR, I think I can even assign you as a reviewer (but I'm not sure).

eharkins (commented):
Thanks @huddlej, this is looking great!

Here's my two review cents:
1. these pages still render:
https://nextstrain--52.org.readthedocs.build/en/52/guides/install/cli-install.html
https://nextstrain--52.org.readthedocs.build/en/52/guides/install/augur_install.html
https://nextstrain--52.org.readthedocs.build/en/52/guides/install/auspice-install.html
even though they aren't linked in any toctree directives, since they are still fetched in https://github.com/nextstrain/docs.nextstrain.org/blob/update-install-docs/src/fetch-docs.py.
Should we remove those from the fetch list and add redirects for them?

2. I got this error when trying to install the conda package in a new environment:
```
MC02T50AUGYGR:ncov eharkins$ conda config --add channels defaults
conda config --add channels bioconda
conda config --add channels conda-forge
Warning: 'defaults' already in 'channels' list, moving to the top
MC02T50AUGYGR:ncov eharkins$ conda config --add channels bioconda
conda config --add channels conda-forge
Warning: 'bioconda' already in 'channels' list, moving to the top
MC02T50AUGYGR:ncov eharkins$ conda config --add channels conda-forge
MC02T50AUGYGR:ncov eharkins$ conda create -n nextstrainconda nextstrain
Solving environment: failed

InvalidVersionSpecError: Invalid version spec: =2.7

MC02T50AUGYGR:ncov eharkins$ conda --version
conda 4.5.9
```
Maybe because I am running an older conda version? Either way the error message isn't super helpful, which made me realize maybe it would be good to add a section for troubleshooting - could just link out to discussion.nextstrain.org or the best place for help with problems installing wherever that is.

davidcroll:
I've just looked at [the #52 draft](https://nextstrain--52.org.readthedocs.build/en/52/install-nextstrain.html) of the install guide.

My comments are:

- good and concise explanation of what nextstrain, augur & auspice is 
- conda installation: The guide at [conda.io](conda.io) is understandable, there's certainly no need to reproduce anything of that on our pages.
- ``sudo apt install docker.io`` failed, packages ``runc`` and ``containerd`` are required, too. Seemingly they don't get installed with ``sudo apt install docker.io`` alone.
- ``nextstrain check-setup --set-default`` may be confusing, as the user has followed installation steps successfully, but still gets some error messages (especially auspice - which has been mentioned before, but seemingly didn't get installed)
- introduce ``docker run hello-world`` to test whether docker is actually running. Here, only following our own guide, it doesn't - which is probably why pillaged the docker website for their installation guide.

I hope I've been helpful again!

huddlej:
Thank you both for the reviews, @eharkins and @davidcroll! I've responded to both of your comments inline below.

> these pages still render: _[...snip...]_ Should we remove those from the fetch list and add redirects for them?

I actually wanted to discuss this with you, @eharkins, and forgot to mention it above. I would prefer to remove these, but is there a reason y'all wanted to copy these into the main docs repo in the first place instead of using the intersphinx functionality? If there was a reason to unify the install docs in one place, we can leave them as they are.

> I got this error when trying to install the conda package in a new environment: _[...snip...]_

Bummer! This outcome might be the norm instead of the exception. I bet updating Conda first will fix this issue. I forgot that we had this same issue with Augur and had to add the following lines at the beginning of [the installation guide](https://docs.nextstrain.org/projects/augur/en/stable/installation/installation.html#using-conda):

```
conda activate base
conda update conda
```

If you run those commands and run the `conda create` command again, does that work, @eharkins?

I just experimented with the `--quiet` flag for `conda config`, to see if I could turn off those annoying warnings about channels and I could not. We can use a more verbose installation command that would avoid the channel commands (and warnings) like so:

```
conda create -n nextstrain -c conda-forge -c bioconda nextstrain
```

> `sudo apt install docker.io` failed, packages `runc` and `containerd` are required, too. Seemingly they don't get installed with `sudo apt install docker.io` alone.

Great catch, @davidcroll. [Docker's own installation instructions](https://docs.docker.com/engine/install/) are always going to be more current (and comprehensive!) than any copy we make. Based on your experience, I'd prefer to replace our current Docker-per-OS tabs with a link to the Docker installation guide. This approach mirrors what we already do for Miniconda by linking users to those installation instructions.

> `nextstrain check-setup --set-default` may be confusing, as the user has followed installation steps correctly, but still gets some error messages (especially auspice - which has been mentioned before, but seemingly doesn't get installed)

Oh no! Auspice should definitely be installed by that point. If you don't mind trying this again to see if it fixes the Auspice installation, would you try destroying your Conda environment and creating a fresh one after updating Conda itself?

```bash
# Update Conda
conda activate base
conda update conda

# Remove the old Nextstrain environment.
conda env remove -n nextstrain

# Create a new Nextstrain environment.
conda create -n nextstrain -c conda-forge -c bioconda nextstrain
```

Related to this point, I wonder whether we should include a "troubleshooting" section that includes exactly the above commands, so users can quickly destroy and rebuild their environment. This seems to be the best way to resolve many Conda environment issues...

For reference, here is the output I get from `nextstrain check-setup --set-default` after I create the Conda environment (and do not have Docker running).

```
nextstrain-cli is up to date!

Testing your setup…
docker: Cannot connect to the Docker daemon at unix:///var/run/docker.sock. Is the docker daemon running?.
See 'docker run --help'.

# docker is not supported
✔ yes: docker is installed
✘ no: docker run works
? unknown: containers have access to >2 GiB of memory
✔ yes: image is new enough for this CLI version

# native is supported
✔ yes: snakemake is installed
✔ yes: augur is installed
✔ yes: auspice is installed

# aws-batch is supported
✔ yes: job description "nextstrain-job" exists
✔ yes: job queue "nextstrain-job-queue" exists
✔ yes: S3 bucket "nextstrain-jobs" exists

All good!  Supported Nextstrain environments: native, aws-batch

Setting default environment to native.
```

> introduce `docker run hello-world` to test whether docker is actually running.

@davidcroll, do you think we need this, if we link to Docker's own installation guide (as mentioned above) instead of trying to recreate a version of that guide here? The Docker guide includes the "hello world" command, so users who follow that guide should be able to confirm Docker's working as expected.

huddlej:
a3be735 implements some of the changes mentioned above including:

 - adding a step to update Conda before installation
 - simplifying the initial Conda environment setup by removing channel configuration
 - providing expected output from the Nextstrain CLI's `check-setup` command
 - replacing custom Docker installation instructions with a link to the official installation guide
 - adding a section on how to troubleshoot a broken installation

eharkins:
> these pages still render: [...snip...] Should we remove those from the fetch list and add redirects for them?

>> I actually wanted to discuss this with you, @eharkins, and forgot to mention it above. I would prefer to remove these, but is there a reason y'all wanted to copy these into the main docs repo in the first place instead of using the intersphinx functionality? If there was a reason to unify the install docs in one place, we can leave them as they are.

I think copying them all in one place was just the simplest thing at the time of docs reorganization, so that they would be easier to find. But now that you have put all this thought into what we want the default user experience to be when installing, I think those specific installation docs will only distract when placed alongside what you have here and are better living in their respective subprojects to be linked to for certain cases.

Thanks for adding troubleshooting section; I get the same error when trying to `conda update conda`, but without trying to run that first I might have continued to blame the issue on installing Nextstrain! Now I know it's probably an issue with my conda installation.

huddlej:
Thanks, @eharkins! I've removed the project-specific install docs from the fetch script and added redirects to the site.

I think I've addressed all of the major concerns that y'all identified, so unless you have any other changes you'd like to see, @davidcroll or @eharkins, I can merge this PR.

tsibley:
> introduce `docker run hello-world` to test whether docker is actually running.

FWIW, `nextstrain check-setup` already runs this, so I don't think this is a necessary detail for most users.

environment.yml
```diff
@@ -11,3 +11,4 @@ dependencies:
     - nextstrain-sphinx-theme>=2020.2
     - sphinx-markdown-tables
     - sphinx-argparse
+    - sphinx-tabs
```
tsibley, line 14: Tabs extension was added in an early commit but then its usage was removed in a subsequent commit. I believe it's no longer necessary?

tsibley (commented):
Overall these new install docs seem really excellent, thank you for putting them together @huddlej!

The flow stood out to me as a bit weird once it hits Docker. As it stands, to use Docker, this doc says to:

1. Install Conda
2. Install all the Nextstrain components into a Conda env
3. Configure `nextstrain` to use the Conda env
4. Install Docker
5. Finally, switch `nextstrain` to use Docker

For updates, it later says:

1. Update the Conda env
2. Update the Docker image

It's a little weird to have the double environments as the recommended setup. Did we talk about forking the path sooner? (I can't recall.)

src/install-nextstrain.rst
```diff
@@ -0,0 +1,159 @@
...
+
+.. code-block:: bash
+
+    conda install -n base -c conda-forge mamba
+
+Then, use Mamba to create the Nextstrain environment.
+
+.. code-block:: bash
+
+    mamba create -n nextstrain -c conda-forge -c bioconda nextstrain
```
tsibley, line 159: I would probably drop mention of Mamba entirely unless we expect its use to be necessary often? People who are familiar with Mamba will already know they can use it without further instruction.

huddlej: I'd like to keep this. I don't see Mamba going away soon (unless the Conda team pulls in its functionality) and it's now the default backend for Snakemake's Conda environment manager (as of [v6.1.0](https://snakemake.readthedocs.io/en/stable/project_info/history.html#id1)). I personally would not have found out about Mamba unless Snakemake mentioned it in the docs, so I would not assume everyone who spends time with Conda knows about it already.

By including its description in the troubleshooting section, we aren't pushing for Mamba as the recommended default (like Snakemake does), but at least we can mention it as a tool used by our team.

huddlej:
> Overall these new install docs seem really excellent, thank you for putting them together @huddlej!

Thank you for reading through them!

> It's a little weird to have the double environments as the recommended setup. Did we talk about forking the path sooner? (I can't recall.)

It's pretty weird. :) We did talk about forking earlier, but the issue kept coming back to how one bootstraps the installation of the CLI. Since our recommended way to install Python and manage workflow environments relies on Conda, we ask users to install the CLI with the Conda interface. We had talked about alternately supporting a standalone binary of the CLI that users could download and run without any additional configuration, but that was going to be a lot more work.

If we wanted to ignore how users get Python installed (assume they already have an appropriate version installed globally), we could split out the CLI/Docker instructions with the `python3 -m pip install nextstrain-cli`, install Docker, `nextstrain update` steps.

It's a weird situation, too, because once you have everything installed via Conda, what is the reason to use the Docker image? I've been thinking for the longest time that you'd want to use the Docker image if you don't want to mess with managing your own environment. But the sense I get from users who mention the Docker image is that they want to run their workflows in the  cloud with their own workflow manager. These are closer to so-called "power users" as opposed to novice users.

It would be nice to discuss all of this stuff in person at some point...it's a complicated topic to distill into GitHub comments. :-/

huddlej:
@tsibley, after talking about these docs today with @jameshadfield, I took a shot at forking the instructions between Conda and Docker a little earlier than before. I still have installation of Conda as a required first step before either approach because we need the user's Python environment to be up-to-date. However, the new instructions for Docker are simpler in that they only install the CLI and Git.

If we get to the point where we can even more easily bootstrap the Python environment and CLI install into a single command without Conda, we can update this structure. Otherwise, does this seem like a reasonable direction?

huddlej:
I found what I think is a replicable disadvantage to using `conda` instead of `mamba`. I wanted to setup an environment with an older Augur installation (11.3.0) so I could test updates to the newer version (12.0.0). To do this, I ran the following command.

```
conda create --yes -n test-nextstrain -c conda-forge -c bioconda nextstrain augur=11.3.0
```

This command took 7 minutes and 28 seconds (448 seconds total) to complete with a lot of time spent “Solving environment”.

Then I ran the update command from inside the environment.

```
conda activate test-nextstrain
conda update nextstrain
```

The last command failed to detect the later Augur package. Next, I tried updating the Nextstrain package and its dependencies:

```
conda update --update-deps nextstrain
```

This command ran for a while (2 min 43 sec or 163 seconds total), with Conda getting confused about the potential paths to solve yarl, google-cloud-storage, and networkx packages. It did eventually finish and detected an updated Augur package. 🎉  This result tells me that we do not need to pin Augur and other package versions in the `nextstrain` meta package for updates to work in the Conda environment; we just need to use the appropriate `--update-deps` flag.

To see if there is any difference in efficiency of the Mamba solver or installer, I repeated the commands above with `mamba` instead of `conda`. The following command finished in 1 min and 39 seconds (99 seconds total or 22% of the time required by `conda`) with little time spent “Solving environment”:

```
mamba create --yes -n test-nextstrain -c conda-forge -c bioconda nextstrain augur=11.3.0
```

Next, I tried to update the environment without specifying dependencies. As with the `conda` command, `mamba` couldn’t find anything to update with this command.

```
mamba update nextstrain
```

 Then I ran the same command with dependency updates requested:

```
mamba update --update-deps nextstrain
```

This did exactly what I wanted, only finding the Augur update and finished in 11 seconds or 7% of the time required by `conda` to do the same thing.

The relative speed-up of the Mamba-based installation and updates was consistent even after cleaning my Conda cache and packages with `conda clean --all` and running the above commands in the opposite order (`mamba` then `conda`, deleting the environment in between these).

Another related and potentially replicable issue is trying to create an environment with only Snakemake installed. The following command consistently tries to install a very old version of Snakemake (~3.X) when the most recent version is 6.1.1.

```
conda create -n snakemake -c conda-forge -c bioconda snakemake
```

However, the following version of the command using `mamba` works as expected, installing version 6.1.1.

```
mamba create -n snakemake -c conda-forge -c bioconda snakemake
```

jameshadfield:
Thanks @huddlej for pushing this!

(The PR RTD build seemed to have disappeared, so I enabled https://docs.nextstrain.org/en/update-install-docs/install-nextstrain.html -- please remove if this wasn't the correct way to do things). 

The install docs have all the information, and I like the conda vs docker split, however it's not clear when reading that these sections should be read as "do one of these, probably not both". This would be the perfect place for [sphinx tabs](https://sphinx-tabs.readthedocs.io/en/latest/), which we previously experimented with here (or so I thought).

I would propose a flow:
* Introduction, with general details
* Install. 3 tabs: "Using Conda", "Using Docker", "From source". The docker section would mention that you can use a conda env, or let people manage their own python install as I imagine those who are comfortable with docker often have their own preferred way of managing python. The "From source" tab would be links to the individual repos (augur, auspice, cli etc).
* Upgrade: Maintain the same 3 tabs.
* Troubleshoot.

Either way, this page is a big improvement over the current fragmented install docs 💯 

Re: conda vs mamba: I haven't used the latter, but the speed increases look significant so I'd be in favour of keeping it in the "troubleshooting" section as you have here

huddlej:
Thank you, @jameshadfield! It definitely wasn't clear that Conda and Docker were mutually exclusive. I've added the tabs package back to the Sphinx environment and config and used it to create linked ("grouped") tabs for the installation and update sections.

I didn't add a tab for installing/updating from source because I realized our "install from source" docs are not complete. For example, the "[Contribute to Augur](https://docs.nextstrain.org/en/latest/guides/contribute/augur_develop.html)" page doesn't explain how to actually get the source code! All three of the main packages are missing this key information. This is obviously a product of sourcing the contributor docs from those repos where the original context made the location of the source code clear.

Instead of delaying this PR much longer, we can continue to refine those developer docs in separate PRs and then update this install guide later.

For the RTD PR build, I've never tried deploying a build for a specific branch, but the GitHub checks for this repo create a corresponding RTD site. If you click on "Show all checks" and click "Details" for the RTD check, that should take you to the PR's docs build.

tsibley:
Thanks for all this work. It's definitely an improvement to what we have currently! I agree that Mamba's gains seem worth pointing out.

I still think the install and subsequent usage flow is a bit weird for the CLI path. The point of the CLI + Docker is a) _not_ having to manage separate environments like those created by Conda and b) being able to run _exactly_ the same environment locally on your computer as remotely in the cloud (Conda does not provide this out of the box). Having to `conda activate nextstrain` before running the `nextstrain` command sort of defeats much of the point of (a). When used with Docker, the `nextstrain` command _is_ the environment manager, so adding Conda adds another confusing layer, in my view. I would rather we document how to install the CLI with `pip` or `pipx` (which later could be simplified by a single-file download-and-run solution).

It's probably worth a higher bandwidth, lower latency face-to-face discussion of this amongst interested folks?

huddlej:
> I would rather we document how to install the CLI with pip or pipx (which later could be simplified by a single-file download-and-run solution).

That sounds reasonable. Would you mind drafting the content for this in this branch?

> It's probably worth a higher bandwidth, lower latency face-to-face discussion of this amongst interested folks?

Definitely a good idea. This seems relevant for our upcoming PGI meeting, since we should all be there...

huddlej:
We still need to discuss these docs in person, as @tsibley mentioned above, and we haven't had time to do that fully. However, I don't want to delay the beneficial revisions in this PR, since they do work and are an improvement on the existing docs. The issues that warrant future discussion (and future GitHub issues to be resolved) are:

1. Whether to use the `nextstrain` meta package with pinned versions for dependencies or do without this meta package completely. As described in the commit message of 4fb61d9, these appear to be our two choices, since not pinning versions causes the current dependency solvers (both Conda's and Mamba's) to produce suboptimal (unacceptable) installations.

2. How/whether to bootstrap a Python environment for users at the minimum required Python for the Nextstrain CLI. I see that using Conda in the current docs appears to defeat the point of the CLI's environment management, but there are actually two separate issues here: 1) managing an environment for the CLI and 2) managing an environment for Nextstrain workflows. The CLI handles the second case, but it does not manage its own installation environment in the first case.

Until the CLI is a standalone binary that does not depend on a Python installation, we need a way to bootstrap the user's environment to a minimum acceptable Python version. We cannot always depend on their system-wide Python to be recent (my own is 2.7.16). Using pip or pipx does not seem to avoid that bootstrapping effort. To use pip, users must figure out how to upgrade their system Python to the correct version for the CLI (I just did this with homebrew to get 3.9.5). To use pipx, Mac users have to install homebrew first to then install pipx. Neither of these are huge asks, but they are also not that different than asking users to install Miniconda. The Conda approach at least appears to provide a similar user experience on Linux and Mac systems.

As a small technical note, we can avoid the requirement of activating a `nextstrain` environment by installing the CLI into the `base` Conda environment that gets activated automatically when you open a new terminal window. That said, I’m open to whatever works best for the most users and think we should test these scenarios with users from our own group and outside it.

PR #53, opened 2021-03-15
Add guide on how to estimate clade frequencies

huddlej:

Adds the Jupyter notebook and corresponding restructured text version of a how-to guide to estimate clade frequencies from SARS-CoV-2 data.

An open question with this guide (and others like it in the future) is where we should source the data. The benefit to the current approach is that it does not require users to prepare any data in advance; data are fetched from the live Nextstrain builds. The disadvantages of this approach are that the guide's static figures quickly diverge from current data and we don't show users how to load their own local data which may be much more relevant.

### Testing

The initial guide was tested by @kistlerk and this version reflects edits based on (most of) her comments. One comment I did not address here was a suggestion to allow users to source their own local data for the guide instead of fetching the live Nextstrain data (see discussion above).


src/guides/bioinformatics/estimate-clade-frequencies.rst
```diff
@@ -0,0 +1,758 @@
...
+------------
+
+Define the URLs to a phylogeny of a global Nextstrain analysis of
+SARS-CoV-2 and the precalculated frequencies for each strain in that
+phylogeny.
+
+.. code:: ipython3
+
+    tree_url = "https://data.nextstrain.org/ncov_global.json"
+    frequencies_url = "https://data.nextstrain.org/ncov_global_tip-frequencies.json"
```
victorlin, line 74:
```suggestion
    tree_url = "https://data.nextstrain.org/ncov_open_global.json"
    frequencies_url = "https://data.nextstrain.org/ncov_open_global_tip-frequencies.json"
```

src/guides/bioinformatics/estimate-clade-frequencies.rst
```diff
@@ -0,0 +1,758 @@
+========================================================
+Estimate frequencies of phylogenetic clades through time
```
victorlin, line 2: > Another potential issue is how we should maintain guides like this that we generate directly from a notebook environment.

What about an extension like [nbsphinx or MyST-NB](https://docs.readthedocs.io/en/stable/guides/jupyter.html)? We'd still have to run the notebook to generate new plots but less manual work than maintaining the same content in both `.ipynb` and `.rst`.

huddlej:
During issue triage we also realized that this guide can be updated to use Nextstrain open data.

PR #60, opened 2021-07-07
Document how to delete files from Groups

huddlej:

Adds a section about how to remove files to the Nextstrain Groups documentation with examples for removing specific files or a group of files matching a pattern.

Separately, expands the previous reference of `nextstrain remote ls` to `nextstrain remote list` for consistency throughout this page.

## Testing

Docs were tested locally and [via RTD hook with this PR](https://nextstrain--60.org.readthedocs.build/en/60/guides/share/nextstrain-groups.html#remove-files-from-your-group).

### DIFF
--- a/src/guides/share/nextstrain-groups.rst
+++ b/src/guides/share/nextstrain-groups.rst
@@ -55,7 +55,8 @@ Define your credentials in this file like so, replacing the “…” values wit
 
 Save this file and return to the command line.
 
-Confirm that you have access to your Nextstrain Groups AWS resources, by listing the contents of your group’s S3 bucket. Replace ``<group>`` below with your group name.
+Confirm that you have access to your Nextstrain Groups AWS resources, by listing the contents of your group’s S3 bucket with :doc:`the nextstrain remote list command <cli:commands/remote/list>`.
+Replace ``<group>`` below with your group name.
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


huddlej:
@tsibley I had never used the `nextstrain remote delete` command before, but I did my best to document how a public health lab member, et al. would use this command to work with their groups S3 bucket. Do you have any words of wisdom/caution about the way I've documented this command here?

src/guides/share/nextstrain-groups.rst
```diff
@@ -126,6 +126,27 @@ Alternately, upload multiple build files at once with wildcard syntax.
...
+============================
+
+You can remove specific files from your group's S3 bucket using the ``nextstrain remote delete`` command.
+For example, the following command removes your group logo and overview files.
+
+.. code-block:: bash
+
+   nextstrain remote delete \
+     s3://nextstrain-<group>/group-logo.png \
+     s3://nextstrain-<group>/group-overview.md
```
tsibley, line 139: `delete` only takes one remote path at a time. This could clearly be improved¹, but right now you'd need two invocations to delete those two objects.

¹ I think the improvement isn't allowing what's suggested here, but instead something like `nextstrain remote delete s3://nextstrain-<group> group-logo.png group-overview.md`, or once we have https://github.com/nextstrain/cli/issues/90, `nextstrain remote delete nextstrain.org/groups/<group> group-logo.png group-overview.md`.

tsibley (changes requested):
@huddlej Thanks for adding this documentation. I added a usage note below.

Separately, I might also suggest linking to the reference docs for `nextstrain remote delete` and other commands mentioned in this document.

src/guides/share/nextstrain-groups.rst
huddlej: Thank you for reviewing this, @tsibley. I'm sorry it took so long to get back to, but 694404d reflects the correction to the remote delete command usage. Separately, 25ea448, links to the corresponding CLI commands docs.

PR #69, opened 2021-08-16
Link to nextclade web tutorial

eharkins:
As @huddlej put it on slack:
> In addition to listing Nextclade under reference guides, it would be excellent to start integrating Nextclade more into the main docs through tutorials and how-to guides. For some users, Nextclade may be the only part of Nextstrain they need to do their job. If they had a link to “Evaluate the quality of your consensus sequences with Nextclade” as one of the first how-to guides, they would be set without needing to know that Nextclade was initially developed as a separate tool

> [This page](https://docs.nextstrain.org/projects/nextclade/en/latest/user/nextclade-web.html), for example, could be linked with a slightly more explanatory title as one of the first links in the main Nextstrain docs tutorials section

This adds that link to better surface the Nextclade tool. The only other currently existing nextclade docs that I could see being "surfaced" on docs.nextstrain.org are:
- [Nextclade CLI](https://docs.nextstrain.org/projects/nextclade/en/latest/user/nextclade-cli.html)
- [Nextalign CLI](https://docs.nextstrain.org/projects/nextclade/en/latest/user/nextalign-cli.html)

But since I couldn't come up with a good place to put them or the words to use in linking to them, I left them out for now.

I linked directly to nextclade docs instead of fetching them based on the assumption that users clicking on the nextclade tutorial would be ok/happy with getting shifted over to nextclade subproject's docs.

### DIFF
--- a/src/index.rst
+++ b/src/index.rst
@@ -9,8 +9,13 @@ Welcome to Nextstrain's documentation!
 
 Nextstrain is an open-source project to harness the scientific and public health potential of pathogen genome data.
 We provide a continually-updated view of publicly available data with powerful analyses and visualizations showing pathogen evolution and epidemic spread.
-Our goal is to aid epidemiological understanding and improve outbreak response.
-If you have any questions, or simply want to say hi, please give us a shout at hello@nextstrain.org.
+Our goal is to aid epidemiological understanding and improve outbreak response and we currently maintain the following tools to help you to analyze and visualize your own data:
+ - `Augur <https://docs.nextstrain.org/projects/augur/en/stable/index.html>`_ (phylogenetic analysis)
+ - `Auspice <https://docs.nextstrain.org/projects/auspice/en/stable/index.html>`_ (visualization)
+ - `Nextstrain CLI <https://docs.nextstrain.org/projects/cli/en/stable/index.html>`_ (command line utility)
+ - `Nextclade <https://docs.nextstrain.org/projects/nextclade/en/latest/index.html>`_ (QC, clade assignment, mutation calling)
+
+If you have any questions, or simply want to say hi, please give us a shout at hello@nextstrain.org or post at `discussion.nextstrain.org <https://discussion.nextstrain.org>`_.
 
 .. toctree::
    :maxdepth: 2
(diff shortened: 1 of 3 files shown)


ivan-aksamentov:
Thanks Eli,

It hink that would do and increase discoverability.
I worry the most about the 
> assumption that users clicking on the nextclade tutorial would be ok/happy with getting shifted over to nextclade subproject's docs.
I just want to make sure Nextclade does not hurt the main docs. Jumping between subprojects might be confusing for beginners.


A nitpicking into:
> "Evaluate the quality of your consensus sequences with Nextclade"
Nextclade Web does [7 things](https://docs.nextstrain.org/projects/nextclade/en/latest/user/algorithm/index.html), and it always does them all (you cannot opt-out). Quality control (QC) is one of them, and seems to be not the most popular one, considering feedback from a subset of users who contacted us. The most popular ones are the alignment, clade assignment, phylogenetic placement. Listing all 7 or even these 3 in the link text is not reasonable, so maybe we could make it something more abstract, like "sequence analysis", but perhaps something also to make it distinct from Nextstrain main pipeline? Perhaps "quick & dirty sequence analysis in the browser"? Except a better English word for "dirty" is needed. I am not very good at those.

But really I don't have a strong opinion on either of these.

rneher:
Thanks! I would probably prefer

`Analyze your sequences with Nextclade`

we could stick an "and assign clades" into this.

rneher:
Up for discussion: I thought it would be useful to have links to the 4 main projects and their documentation right at the top of the docs landing page. But fine to roll back if people don't like this.

eharkins:
Thanks for the corrections about what Nextclade does! 

> I just want to make sure Nextclade does not hurt the main docs. Jumping between subprojects might be confusing for beginners.

This will be alleviated in part by https://github.com/nextstrain/nextclade/pull/505 since it makes navigating back to the main docs easier, but you may still be right. Also just a note for me to remember that if we do decide to fetch the nextclade doc and build it into the docs.nextstrain.org project, we will have to change a few of the relative links in that doc to be absolute links so that they work in both contexts.

eharkins:
https://nextstrain--69.org.readthedocs.build/en/69/tutorials/index.html Here is what this looks like as a built version of the docs.

eharkins:
This is ready to merge from my perspective unless there are concerns about adding links to the 4 subprojects on the main docs landing page and/or linking to the nextclade subproject for the tutorial rather than fetching the doc and building it into the main docs.

ivan-aksamentov (approved):

ivan-aksamentov:
LGTM.

We can always add the 

> links to the 4 subprojects on the main docs landing page

and

fetching the doc and building it into the main docs.

later if we decide so.

eharkins:
> links to the 4 subprojects on the main docs landing page

was added here already by @rneher: https://github.com/nextstrain/docs.nextstrain.org/pull/69/files#diff-168e13fab1ec9b038f23cc43479f23a901729ab78e79aec8e9c28f9f2a1fcbb0R13

PR #71, opened 2021-08-26
Rewrite data formats page to cover dataset, sidecar, and node data JSON files

tsibley:
Thinking through our data model for CLI and API work, I found myself
writing down what JSON files we use where in order to help orient
myself.  Based on text first drafted in my personal notes.¹

¹ https://github.com/tsibley/blab-standup/blob/master/2021-08-25.md


tsibley:
Incorporated as part of #75 instead.

PR #72, opened 2021-08-30
Lay of the land

tsibley:
Full rewrite, starting from some (less extensive) notes I wrote up
earlier this year describing the different pieces of Nextstrain to
someone.

The intent is to help people construct a mental model of the Nextstrain
ecosystem so they'll find it easier to navigate the docs and know what
they're looking for.  Each paragraph tries to build upon the previous
one to link together terminology, with hyperlinks to the topics
mentioned for immediate jumping off points and further reading.

### DIFF
--- a/src/index.rst
+++ b/src/index.rst
@@ -13,12 +13,12 @@ understanding and improve outbreak response.
 
 .. note::
 
-    This documentation site aims to cover all parts of Nextstrain.  If you already
-    have your bearings, see the :ref:`toc` below to dive right in.  If you're
-    interested in how to interpret what you see on nextstrain.org, please start
+    This documentation site aims to cover all parts of Nextstrain.  If you
+    already have your bearings, see the :ref:`toc` below to dive right in.  If
+    you're interested in how to interpret what you see on nextstrain.org, start
     with :doc:`/learn/interpret/index`.  If you’re ready to go beyond using the
-    website (or just plain curious!), please continue reading for an overview of
-    how the different pieces of Nextstrain fit together.
+    website (or just plain curious!), continue reading for an overview of how
+    the different pieces of Nextstrain fit together.
 
 The core components of Nextstrain are :doc:`Augur <augur:index>` and
 :doc:`Auspice <auspice:index>`.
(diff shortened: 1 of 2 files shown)


tsibley:
Best way to read this is via the [PR preview](https://nextstrain--72.org.readthedocs.build/en/72/).

trvrb:
I really like this and think it's a big improvement. I didn't have much to add except a small fix to call change "data files" to "dataset files" for consistency.

src/index.rst
```diff
@@ -1,28 +1,228 @@
...
+please `email the Nextstrain team <hello@nextstrain.org>`__.
 
 .. toctree::
    :maxdepth: 2
    :titlesonly:
-   :caption: Table of contents
+   :caption: Table of Contents
+   :name: toc
 
-   Nextstrain Documentation <self>
```
huddlej, line 25: Thank you for removing this! Denisse and I were just discussing how confusing this link is.

src/index.rst
```diff
@@ -1,28 +1,228 @@
...
+visualization and analysis.  Builds run dozens of commands and often use
+`Snakemake <https://snakemake.readthedocs.io>`__ to :doc:`manage the pipeline
+workflow </guides/bioinformatics/augur_snakemake>`, but any workflow system can
+be used, such as `Nextflow <https://nextflow.io>`__ or `WDL
+<https://openwdl.org>`__.  As an example, our core builds are organized as `Git
+repositories <https://git-scm.com>`__ hosted on `GitHub
+<https://github.com/nextstrain>`__ which contain a Snakemake workflow using
+Augur, configuration, and data.
+
+.. graphviz::
```
huddlej, line 143: This embedded graphviz is an excellent standard to set for our documentation. I would love to use this approach to replace [the outdated ncov workflow diagram](https://docs.nextstrain.org/projects/ncov/en/latest/analysis/orientation-workflow.html) now that it's in RTD.

huddlej (approved):
I love this. It's easy to follow and fills a huge gap in the current docs.

My only very minor comment is that we use "please" a lot and could easily drop this bit of happy talk from our docs. I also get that this is a standard in our current docs, so it's not a fault of this PR exactly.

tsibley:
Ah, thanks for pointing out the pleases. I agree it's a filler "happy talk" word in this context. Dropped the uses from this page in 91fd28b.

PR #74, opened 2021-08-31
Add a CI job which builds the docs with fatal warnings

tsibley:
Expected to fail currently as our docs build has warnings.  Next step is
to resolve the warnings and then keep the build warnings clean.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -1,0 +1,24 @@
+name: CI
+
+on:
+  - push
+  - pull_request
+
+jobs:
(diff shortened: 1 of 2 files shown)


tsibley:
Warnings free and CI job is now passing. Let's keep it that way! :-)

Unfortunately due to how we fetch and incorporate files from other repos, I had to make [a small commit to Auspice](https://github.com/nextstrain/auspice/commit/3ff0f755f996cc8aecb7fdeb859d1dd7d6ac4396) to address two of the warnings.

PR #79, opened 2021-09-24
fix: mamba update --all requires explicit inclusion of -c bioconda

corneliusroemer:
If `-c bioconda` is not included, then mamba will only update primary channel packages.


huddlej (approved):
This worked for me! Thank you, @corneliusroemer!

PR #80, opened 2021-11-19
Run fetch-docs.py under the same Python that Sphinx is running under

tsibley:
Otherwise, fetch-docs.py may run under a completely different Python
version/environment/install than Sphinx because fetch-docs.py uses
whatever `python` is (via `#!/usr/bin/env python`) while Sphinx may be
invoked explicitly with a non-default Python.  @victorlin observed the
result of that failure mode in the VSCode integration with Sphinx.

### DIFF
--- a/src/conf.py
+++ b/src/conf.py
@@ -17,7 +17,7 @@ sys.path.insert(0, os.path.abspath('.'))
 # -- Fetch docs --------------------------------------------------------------
 
 import subprocess
-subprocess.check_call(['./fetch-docs.py'])
+subprocess.check_call([sys.executable or "python", './fetch-docs.py'])
 
 # -- Project information -----------------------------------------------------
 


victorlin (approved):
Neat, that works!

1. Do we still need `#!/usr/bin/env python` then?
2. Would `sys.executable` ever be empty in our use case (do we need `or "python3"`)?

tsibley:
1. I'd prefer to leave it for times when we manually run the program, such as when testing modifications to it. Minor thing, but nice not to have to remember to type `python`.
2. Don't know if it'll ever be empty in practice for us (probably not!), but it's documented that it can be, so easy to fallback to "whatever we got" instead of breaking. Maybe should use `python` instead of `python3`, I just used the latter out of habit.

tsibley:
Updated to use `python` to match the shebang and threw a cross-link to our Slack convo into the commit msg for future ref.

victorlin (approved):

PR #81, opened 2021-11-19
Update installation instructions for macOS/Windows/WSL

victorlin:
Preview: https://nextstrain--81.org.readthedocs.build/en/81/install.html

After experimenting with Nextstrain installation on different operating systems, there were notable differences that warranted sectioning the guide per OS.

This change rewrites the installation steps with a grouping of tabs per tested OS.

A new Background section is added before installation steps, retaining notable context from the current documentation.

### DIFF
--- a/src/install.rst
+++ b/src/install.rst
@@ -126,6 +126,12 @@ These instructions will install the Nextstrain CLI and tools to run and view you
             conda activate nextstrain
 
       6. Install the remaining Nextstrain components by `installing Docker Desktop with WSL 2 backend <https://docs.docker.com/desktop/windows/wsl/>`_.
+
+         .. note::
+
+            You may have to restart your machine when configuring WSL (Windows Subsystem for Linux).
+            If so, remember to open a new Anaconda PowerShell Prompt and run ``conda activate nextstrain`` before the next step.
+
       7. Confirm that the installation worked.
 
          .. code-block:: none
--- a/src/install.rst
+++ b/src/install.rst
@@ -25,7 +25,7 @@ Background
 ==========
 
 `Conda <https://docs.conda.io/en/latest/>`_ is a package and environment management system that allows you to install Python and other software into controlled environments without disrupting other software you have installed (e.g., on your computer, your shared cluster, etc.).
-Conda provides an appropriate version of Python required by all approaches to installing Nextstrain tools.
+Conda provides an appropriate version of Python required by all approaches to installing Nextstrain tools. Miniconda is the minimal installation of the command-line interface to conda.
 
 `Mamba <https://github.com/mamba-org/mamba>`_ is a drop-in replacement for most ``conda`` functionality that implements a faster dependency solving algorithm in C++ and multithreaded downloads.
 As a result, mamba can install conda packages much faster and more accurately than the original conda installer.
@@ -46,7 +46,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
          If you are an experienced user, you can replace ``conda`` with ``pip`` but :doc:`note the extra installation steps for augur <augur:installation/installation>` and :doc:`install auspice via npm <auspice:introduction/install>`.
 
-      1. `Install Miniconda <https://docs.conda.io/projects/conda/en/latest/user-guide/install/download.html>`_.
+      1. `Install Miniconda <https://docs.conda.io/en/latest/miniconda.html>`_.
       2. Open a terminal window.
       3. Install mamba on the ``base`` conda environment:
 
@@ -89,7 +89,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
          Due to installation constraints, there is no way to use the native build/view environment on Windows directly. Follow steps for **WSL on Windows** if the native environment is desired.
 
-      1. `Install Miniconda <https://docs.conda.io/projects/conda/en/latest/user-guide/install/download.html>`_.
+      1. `Install Miniconda <https://docs.conda.io/en/latest/miniconda.html>`_.
       2. Install Visual C++ build tools following `this guide <https://stackoverflow.com/a/64262038>`_.
 
          - This is necessary for a dependency of Nextstrain CLI and `we are investigating options to eliminate this step <https://github.com/nextstrain/cli/issues/31#issuecomment-970641263>`_.
--- a/src/install.rst
+++ b/src/install.rst
@@ -60,7 +60,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
             .. code-block:: bash
 
-                  conda install -c conda-forge -c bioconda augur auspice snakemake --yes
+               conda install -c conda-forge -c bioconda augur auspice snakemake --yes
 
       5. Confirm that the installation worked.
 
@@ -141,7 +141,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
             .. code-block:: bash
 
-                  conda install -c conda-forge -c bioconda augur auspice snakemake --yes
+               conda install -c conda-forge -c bioconda augur auspice snakemake --yes
 
       6. Confirm that the installation worked.
 
--- a/src/install.rst
+++ b/src/install.rst
@@ -33,14 +33,7 @@ When you use the Docker build/view environment, you don’t need to manage any o
 Installation Steps
 ==================
 
-There are two components to install and configure. These are all the possible configuration options, with availability varying by operating system:
-
-1. Nextstrain CLI
-2. Nextstrain build/view environment:
-
-   a. Docker (build + view)
-   b. Native (build + view)
-   c. AWS Batch (build) + Docker/Native (view)
+These instructions will install the Nextstrain CLI and tools to run and view your own Nextstrain analyses. Configuration options vary by operating system.
 
 .. tabs::
 
@@ -60,7 +53,7 @@ There are two components to install and configure. These are all the possible co
             conda create -n nextstrain -c bioconda nextstrain-cli --yes
             conda activate nextstrain
 
-      4. Install the Nextstrain build/view environment. There are two options:
+      4. Install the remaining Nextstrain components. There are two options:
 
          a. Docker (recommended) – install Docker Desktop using `the official guide <https://docs.docker.com/desktop/mac/install/>`_.
          b. Native – install all the necessary software using conda:
@@ -75,7 +68,7 @@ There are two components to install and configure. These are all the possible co
 
             nextstrain check-setup --set-default
 
-         The final output from the last command should look like this, where ``<option>`` is the environment chosen in the previous step:
+         The final output from the last command should look like this, where ``<option>`` is the option chosen in the previous step:
 
          .. code-block:: none
 
@@ -101,7 +94,7 @@ There are two components to install and configure. These are all the possible co
             conda create -n nextstrain -c bioconda nextstrain-cli --yes
             conda activate nextstrain
 
-      5. Install the Nextstrain build/view environment by `installing Docker Desktop with WSL 2 backend <https://docs.docker.com/desktop/windows/wsl/>`_.
+      5. Install the remaining Nextstrain components by `installing Docker Desktop with WSL 2 backend <https://docs.docker.com/desktop/windows/wsl/>`_.
       6. Confirm that the installation worked.
 
          .. code-block:: none
@@ -138,7 +131,7 @@ There are two components to install and configure. These are all the possible co
             conda create -n nextstrain -c bioconda nextstrain-cli --yes
             conda activate nextstrain
 
-      5. Install the Nextstrain build/view environment. There are two options:
+      5. Install the remaining Nextstrain components. There are two options:
 
          a. Docker (recommended) – on Windows, `install Docker Desktop for WSL 2 backend <https://docs.docker.com/desktop/windows/wsl/>`_.
 
@@ -156,13 +149,13 @@ There are two components to install and configure. These are all the possible co
 
             nextstrain check-setup --set-default
 
-         The final output from the last command should look like this, where ``<option>`` is the environment chosen in the previous step:
+         The final output from the last command should look like this, where ``<option>`` is the option chosen in the previous step:
 
          .. code-block:: none
 
             Setting default environment to <option>.
 
-Optionally, :doc:`configure the AWS Batch build environment <cli:aws-batch>` to run the build step on AWS.
+Optionally, :doc:`configure AWS Batch <cli:aws-batch>` if you'd like to run ``nextstrain build`` on AWS.
 
 Next, try :doc:`tutorials/quickstart`.
 
--- a/src/install.rst
+++ b/src/install.rst
@@ -74,7 +74,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
             .. code-block:: bash
 
-               mamba install -c conda-forge -c bioconda augur auspice snakemake --yes
+               mamba install -c conda-forge -c bioconda augur auspice nextalign snakemake git --yes
 
       6. Confirm that the installation worked.
 
@@ -185,7 +185,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
             .. code-block:: bash
 
-               mamba install -c conda-forge -c bioconda augur auspice snakemake --yes
+               mamba install -c conda-forge -c bioconda augur auspice nextalign snakemake git --yes
 
       7. Confirm that the installation worked.
 
@@ -257,7 +257,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
             .. code-block:: bash
 
-                  mamba install -c conda-forge -c bioconda augur auspice snakemake --yes
+                  mamba install -c conda-forge -c bioconda augur auspice nextalign snakemake git --yes
 
       5. Confirm that the installation worked.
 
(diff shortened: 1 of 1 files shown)


src/install.rst
```diff
@@ -6,148 +6,184 @@ Installing Nextstrain
...
-    .. code-block:: bash
+         .. code-block:: none
 
-        Setting default environment to docker.
+            conda create -n nextstrain -c bioconda nextstrain-cli --yes
+            conda activate nextstrain
 
-    Finally, download the latest Docker image for Nextstrain.
+      5. Install the Nextstrain build/view environment by `installing Docker Desktop with WSL 2 backend <https://docs.docker.com/desktop/windows/wsl/>`_.
+      6. Confirm that the installation worked.
```
huddlej, line 105: I'm curious about your experience with this, @victorlin, but in my efforts to install Docker Desktop and WSL 2, I had to restart my computer at least once (after installing the Linux kernel update prerequisite for Docker). Was that the case for you, too?

If so, we should warn the reader that they will need to restart their computer during Nextstrain installation (so it doesn't come as an annoying surprise halfway through installing our dependencies) and also consider adding a reminder to open a new Anaconda prompt before this "Confirm that the installation worked" step.

It's actually worth clearly stating for users in any case that they need to run Nextstrain from an Anaconda prompt.

src/install.rst
```diff
@@ -6,148 +6,184 @@ Installing Nextstrain
...
 
-    .. code-block:: bash
+         .. code-block:: none
 
-        Setting default environment to docker.
+            conda create -n nextstrain -c bioconda nextstrain-cli --yes
+            conda activate nextstrain
 
-    Finally, download the latest Docker image for Nextstrain.
+      5. Install the Nextstrain build/view environment by `installing Docker Desktop with WSL 2 backend <https://docs.docker.com/desktop/windows/wsl/>`_.
```
huddlej, line 104: Both Docker and Microsoft documentation sites are written for highly technical users. I know all of the details our users will need to install Docker are in those docs, for example, but just parsing those docs could be painful for someone without much technical experience (i.e., our Windows users).

I don't know how much we want to go down this path, but one option to alleviate this pain would be for us to provide our own parallel installation guide with screenshots and less technical prose. I'm not recommending that approach for this PR, but we should see how complicated our users actually find this experience and keep an open mind about going this direction.

src/install.rst
```diff
@@ -6,148 +6,184 @@ Installing Nextstrain
...
 
-      .. code-block:: bash
+            conda create -n nextstrain -c bioconda nextstrain-cli --yes
+            conda activate nextstrain
 
-         conda update -n base conda
+      5. Install the Nextstrain build/view environment. There are two options:
 
-      Update the Nextstrain environment.
+         a. Docker (recommended) – on Windows, `install Docker Desktop for WSL 2 backend <https://docs.docker.com/desktop/windows/wsl/>`_.
```
huddlej, line 142: I hadn't thought about this combination of possible installation options and I haven't heard of anyone using it before.

Is it worth keeping a separate "WSL for Windows" option, if the "Windows" option does everything we need it to? I guess the main argument for this option is to provide a way for users to install with Conda (and avoid Docker). @emmahodcroft, do you have any opinions about this?

Either way, these are great instructions for a "Linux" tab (minus the Windows-specific bits).

src/install.rst
```diff
@@ -6,148 +6,184 @@ Installing Nextstrain
...
 
-    Finally, download the latest Docker image for Nextstrain.
+      5. Install the Nextstrain build/view environment by `installing Docker Desktop with WSL 2 backend <https://docs.docker.com/desktop/windows/wsl/>`_.
+      6. Confirm that the installation worked.
 
-    .. code-block:: bash
+         .. code-block:: none
 
-        nextstrain update
+            nextstrain check-setup --set-default
```
huddlej, line 109: When I tried this on a fresh Windows and Docker installation, this command failed while trying to pull the “hello-world” image with an error message about not having permission to access the image. Even though Docker Desktop was running, something was still broken. I opened the Docker dashboard and found a notice saying that I should login to have access to my images, etc. I logged in and ran the `check-setup` command again and it worked.

Did you have a similar experience, @victorlin, where you needed to login to Docker? Were you already logged in as a matter of habit?

huddlej (changes requested):
Thank you for putting this together, @victorlin! It's a lot of work covering all of these possible paths to installation, but these docs seem much clearer to me than the previous version.

My biggest request is for the addition of a "Linux" tab in the OS-specific instructions. I imagine these instructions will largely look the same as the WSL for Windows.

I'd also love comments from @j23414 and @DSequ about the Windows/Mac docs and any pitfalls they experienced that aren't covered by this version. After we feel good about this version, we should also run it by some users outside of our group. Bryan Tegomoh, for one, has been excited to try out the Windows docs, so his input will be invaluable here (I'll follow up with him separately).

victorlin:
Thanks for the helpful comments and perspective @huddlej! Much of it revolves around clarity for non-technical users, which I hadn't put much focus on. I will push changes and address them individually.

src/install.rst
```diff
@@ -6,148 +6,184 @@ Installing Nextstrain
...
 
-.. tabs::
+      .. note::
 
-   .. group-tab:: Conda
+         If you are an experienced user, you can replace ``conda`` with ``pip`` but :doc:`note the extra installation steps for augur <augur:installation/installation>` and :doc:`install auspice via npm <auspice:introduction/install>`.
 
-      Create a Conda environment named ``nextstrain``.
-      This command will install Nextstrain and its dependencies.
+      1. Install `Anaconda or Miniconda <https://docs.conda.io/projects/conda/en/latest/user-guide/install/download.html>`_.
```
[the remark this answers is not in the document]
j23414: I second the preference for Miniconda, unless there's a reason to mention Anaconda.

src/install.rst
victorlin: That's weird, I don't have a Docker account and never signed into Docker Desktop on macOS or Windows. Nevertheless, I'll look into setting up a new Windows user and going through these steps again for overall reproducibility to address other comments.

src/install.rst
j23414: My impression was that [Singularity](https://www.blopig.com/blog/2021/09/using-singularity-on-windows-with-wsl2/) was easier to install on Windows than Docker (singularity can pull and run Docker images, without requiring sudo. And HPC clusters usually have singularity but no docker). Just a discussion point, focusing on Docker is fine.

tsibley: I'll note that we don't currently support Singularity as a runtime for the CLI, though it's a [long standing todo](https://github.com/nextstrain/cli/issues/2). On Windows, Singularity will still need to run in a VM though, which means managing that directly unless Singularity packages tools to do that automatically (e.g. like Docker Desktop does).

src/install.rst
```diff
@@ -6,148 +6,184 @@ Installing Nextstrain
...
+There are two components to install and configure. These are all the possible configuration options, with availability varying by operating system:
 
-.. code-block:: bash
+1. Nextstrain CLI
+2. Nextstrain build/view environment:
 
-    conda update -n base conda
+   a. Docker (build + view)
+   b. Native (build + view)
+   c. AWS Batch (build) + Docker/Native (view)
```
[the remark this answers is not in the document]
victorlin: Updated to remove references to build/view environment concept per our [Slack discussion](https://bedfordlab.slack.com/archives/C7SDVPBLZ/p1637707993122600): https://github.com/nextstrain/docs.nextstrain.org/pull/81/commits/c79e78ff850760424ace5293b0ac3f50108e3146

src/install.rst
```diff
@@ -6,148 +6,184 @@ Installing Nextstrain
...
 
-Install Nextstrain with Conda or Docker
-=======================================
+.. tabs::
 
-Next, decide whether you prefer to install Nextstrain with Conda or Docker.
-We recommend Conda for M1 Mac and Windows users.
-Docker is not yet ready for widespread use on the M1 Mac.
-Similarly, there are still significant obstacles to running Docker with Windows, as documented in `our issue tracking the problems <https://github.com/nextstrain/cli/issues/31>`_.
+   .. group-tab:: macOS
```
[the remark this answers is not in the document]
victorlin: Added a tab for Ubuntu Linux and tested on an AWS EC2 instance: https://github.com/nextstrain/docs.nextstrain.org/pull/81/commits/79545810999793d277ce2826dc4fd99ec78035c5

src/install.rst
victorlin: I agree, Windows+Docker should be sufficient for most users. But if a Windows user wants to install a specific version of augur/auspice or do any development work, WSL is the only option. For now, we can keep the tab here with the option to move it elsewhere later?

src/install.rst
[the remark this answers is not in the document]
victorlin: Good idea, updated: https://github.com/nextstrain/docs.nextstrain.org/pull/81/commits/747588a9dd81fa2afee73295b12c99fa24f9c8da

src/install.rst
```diff
@@ -6,148 +6,241 @@ Installing Nextstrain
...
+            conda create -n nextstrain -c bioconda nextstrain-cli --yes
+            conda activate nextstrain
+
+      5. Install the remaining Nextstrain components. There are two options:
+
+         a. Docker (recommended) – on Windows, `install Docker Desktop for WSL 2 backend <https://docs.docker.com/desktop/windows/wsl/>`_.
+
+            - Make sure to follow through the last step of enabling **WSL Integration**.
+
+         b. Native – install all the necessary software using conda:
```
victorlin, line 138: Request from @j23414 [on Slack](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1637685730181700?thread_ts=1637626719.175300&cid=C01LCTT7JNN):

> Any chance we could add a "mamba" tab (similar to the conda and docker tabs)? Or simply add mamba instructions to the conda tab ([link to notes](https://github.com/j23414/usingnextstrain#1-nextstrain-installation)).

victorlin: Updated to use mamba for all package installation commands: https://github.com/nextstrain/docs.nextstrain.org/pull/81/commits/af210cd12e62324435590b3c575ce7775aa20a05

src/install.rst
victorlin: I just re-checked and I haven't needed to login to Docker Desktop on any of the tested platforms. Let me know if you run into this again though, it could be beneficial to add in the troubleshooting section.

src/install.rst
victorlin: The Docker and WSL installs were system-wide so I wasn't able to check this on a new user account (and I don't feel like wiping and re-installing Windows yet again). But I vaguely remember restarting at least once for WSL, so I've added that as a note: https://github.com/nextstrain/docs.nextstrain.org/pull/81/commits/0a7153e155c5f77383ce1f90878c1f1c444726aa

src/install.rst
```diff
@@ -6,148 +6,184 @@ Installing Nextstrain
...
-        conda create -n nextstrain -c conda-forge -c bioconda nextstrain-cli git
+      .. note::
 
-    Confirm that the installation worked and configure the CLI to use Docker as the default environment manager.
+         Due to installation constraints, there is no way to use the native build/view environment on Windows directly. Follow steps for **WSL on Windows** if the native environment is desired.
 
-    .. code-block:: bash
+      1. Install `Anaconda or Miniconda <https://docs.conda.io/projects/conda/en/latest/user-guide/install/download.html>`_.
+         Miniconda is sufficient for this guide.
+      2. Install Visual C++ build tools following `this guide <https://stackoverflow.com/a/64262038>`_.
```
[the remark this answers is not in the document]
victorlin: Looks like we will be dropping `netifaces` soon which will eliminate this step for the better. In the meantime, I've copied the steps from the StackOverflow answer to our doc directly: https://github.com/nextstrain/docs.nextstrain.org/pull/81/commits/94beb5c0c0973a012a693ad28119a72c2738c7be

src/install.rst
victorlin: I added explicit step-by-step instructions for Miniconda in https://github.com/nextstrain/docs.nextstrain.org/pull/81/commits/94beb5c0c0973a012a693ad28119a72c2738c7be and started doing the same for Docker/WSL, but didn't include those since it's complex and I don't want to hinder the brevity of our docs.

Will keep this as-is, but a separate installation guides with screenshots as @huddlej mentioned is worth considering later based on user feedback.

src/install.rst
```diff
@@ -6,197 +6,314 @@ Installing Nextstrain
...
 
-         conda activate nextstrain
-         conda update nextstrain-cli
+      .. note::
 
-      Download the latest image with the Nextstrain CLI.
+         If you are an experienced user, you can replace ``conda`` with ``pip`` but :doc:`note the extra installation steps for augur <augur:installation/installation>` and :doc:`install auspice via npm <auspice:introduction/install>`.
 
-      .. code-block:: bash
+      1. `Install WSL 2 <https://docs.microsoft.com/en-us/windows/wsl/install>`_.
```
victorlin, line 153: Might be possible to simplify this step by telling users to download/install something from Microsoft Store. Will need to revisit.

src/install.rst
```diff
@@ -6,197 +6,314 @@ Installing Nextstrain
...
 
-      Update the base Conda environment.
+         .. code-block:: none
 
-      .. code-block:: bash
+            mamba create -n nextstrain -c bioconda nextstrain-cli --yes
+            conda activate nextstrain
 
-         conda update -n base conda
+      6. Install the remaining Nextstrain components by `installing Docker Desktop with WSL 2 backend <https://docs.docker.com/desktop/windows/wsl/>`_.
```
victorlin, line 128: Need to double-check if non-admin user can follow the same steps.

src/install.rst
huddlej: That sounds good!

src/install.rst
```diff
@@ -6,197 +6,314 @@ Installing Nextstrain
...
-         Setting default environment to native.
+         a. Docker (recommended) – install Docker Desktop using `the official guide <https://docs.docker.com/desktop/mac/install/>`_.
+         b. Native – install all the necessary software using mamba:
 
-      Whenever you open a new terminal window to work on a Nextstrain analysis, remember to activate the Nextstrain Conda environment with ``conda activate nextstrain``.
-      Next, try :doc:`tutorials/quickstart`.
+            .. code-block:: bash
 
-   .. group-tab:: Docker
+               mamba install -c conda-forge -c bioconda augur auspice snakemake --yes
```
huddlej, line 77: In our current docs, we ask folks to install:

```
 augur auspice nextstrain-cli nextalign snakemake awscli git pip
```

The `pip` and `awscli` dependencies don't seem to be necessary anymore. Conda used to complain when you didn't explicitly install pip in your environment, but I didn't see this in my recent installation. The Nextstrain CLI's `remote` commands do most of what a Nextstrain user would need from the `awscli`.

We should keep the `nextalign` dependency, since it is in the Dockerfile, used in the ncov work, and is likely to be used in seasonal-flu and other workflows, soon. (There's also a `nextclade` conda package, but I don't know if we plan to use this as broadly yet.)

I'm not sure about dropping the `git` dependency. It's necessary for anyone to run our tutorials (until we document how to download from GitHub, for example) and novice users may not have it installed already or it might be a pain for them. I'd err on the side of keeping it, for now, unless we know that everyone already has it installed.

src/install.rst
```diff
@@ -6,197 +6,314 @@ Installing Nextstrain
...
-As a result, Mamba can install Conda packages much faster and more accurately than the original Conda installer.
+         a. Docker (recommended) – on Windows, `install Docker Desktop for WSL 2 backend <https://docs.docker.com/desktop/windows/wsl/>`_.
+
+            - Make sure to follow through the last step of enabling **WSL Integration**.
+
+         b. Native – install all the necessary software using mamba:
+
+            .. code-block:: bash
+
+               mamba install -c conda-forge -c bioconda augur auspice snakemake --yes
```
huddlej, line 188: See comment above about including nextalign and git here.

src/install.rst
```diff
@@ -6,197 +6,314 @@ Installing Nextstrain
...
+
+               .. code-block:: bash
+
+                  conda activate nextstrain
+
+         b. Native – install all the necessary software using mamba:
+
+            .. code-block:: bash
+
+                  mamba install -c conda-forge -c bioconda augur auspice snakemake --yes
```
huddlej, line 260: See comment above about including nextalign and git here.

huddlej (changes requested):
Thank you for all the updates, @victorlin! This looks really nice. I had one last request about the specific conda packages being installed for the "native" mode, but that's it.

src/install.rst
victorlin: Got it, I haven't tried ncov with native yet and had git preinstalled. Fixed here: https://github.com/nextstrain/docs.nextstrain.org/pull/81/commits/4584bd292626e4c6671d37e5870157369a7557dd

src/install.rst
victorlin: Fixed here: https://github.com/nextstrain/docs.nextstrain.org/pull/81/commits/4584bd292626e4c6671d37e5870157369a7557dd

src/install.rst
victorlin: Fixed here: https://github.com/nextstrain/docs.nextstrain.org/pull/81/commits/4584bd292626e4c6671d37e5870157369a7557dd

huddlej (approved):

src/install.rst
victorlin: Moved to #82

src/install.rst
victorlin: Moved to #82

PR #84, opened 2021-12-21
Update terminology and tutorials

victorlin:
Preview: https://nextstrain--84.org.readthedocs.build/en/84/

Summary:

- Convert Zika tutorial to reST.
- Add glossary page and link to glossary terms with reST syntax.
- Update terminology:
    - An attempt to resolve **workflow**/**pipeline**/**build** ambiguity: a **workflow** bundles one or more related **builds** which each produce a **dataset**.
    - Add a new term **runtime** to replace existing term **build environment**.
- Rename tutorial pages with new terminology
    - Add RTD redirects in new `redirects.yaml` file (currently manually synced).

Page-specific changes:

- Parts of a whole:
    - Add section on workflow term.
    - Improve existing graph visuals for dataset term.
- Running a workflow (formerly quickstart):
    - Remove references to build term - purpose of tutorial is to run a Snakemake workflow.
- Creating a workflow (formerly Zika):
    - Clarify difference between **build** and **workflow**.
    - Improve instructions for Docker runtime.
    - Use `nextstrain view` instead of `auspice view` to visualize the dataset.
- Creating a bacterial pathogen workflow (formerly TB):
    - "ancestral states" -> "ancestral traits" for consistency with other tutorials.


victorlin:
As I mentioned in internal discussions, I've spent some time getting familiar with these terms across current docs and this is what I've gathered:

- `build` is an overloaded term, used to describe at least these 3 distinct concepts:
    1. (n.) a sequence of steps that produces JSON output (e.g. individual builds in `builds.yaml`)
    2. (n.) a collection of the above concept (e.g. the entire ncov "build")
    3. (v.) the action of running the above concept (e.g. `nextstrain build` command, `built with {repo}`)
- `workflow` can also be confusing per this thread. It is used synonymously to `build (a)` and `build (b)`, but mostly the latter (e.g. SARS-CoV-2 workflow).
- `dataset` representing the output of `build (a)` is a well-scoped term (though I think the name itself is confusing since there is already an existing set of data with metadata.tsv+sequences.fasta, but it's ok).
- `build (a)` and `build (b)` are basically the same thing in Zika, but it gets confusing in SARS-CoV-2 where it is pretty distinct.

To address the ambiguity, this PR proposes specific definitions:
- `build` represents 1 noun `(a)` and 1 verb `(c)` since these concepts are already exposed to users through obvious references (`builds.yaml`, `nextstrain build`, `built with {repo}`)
- `workflow` represents `(b)`: a collection of build(s), used by workflow managers with varying implementation. Our core workflows are either single-build (e.g. zika) or mutli-build (e.g. ncov).
This will require some replacement of "build" with workflow" (e.g. maybe "Zika pathogen workflow" instead of "Zika pathogen build").

---

Besides ambiguity, another problem with the existing loosely-defined terms is the extended terms, for example **build script**, where the meaning is specific to implementation (zika tutorial has a singular Snakefile build script; but in ncov, builds are orchestrated through a combination of Snakefile+*.smk+builds.yaml, config, etc.).
This was the biggest struggle for me when going through the tutorials – realizing and understanding the differences between Zika and SARS-CoV-2 workflows, and the fact that these are just two ways of doing it.

src/tutorials/zika.rst
```diff
@@ -0,0 +1,300 @@
...
+.. contents::
+   :local:
+
+Prepare the Sequences
+---------------------
+
+.. admonition:: Sequence data overview
+   :class: dropdown, note
+
+   A :term:`Nextstrain build<build>` typically starts with a collection of pathogen sequences in a single `FASTA <https://en.wikipedia.org/wiki/FASTA_format>`_ file and a corresponding table of metadata describing those sequences in a tab-delimited text file. For this tutorial, we will use example data containing 34 virus sequences.
```
tsibley, line 66: Since this is a tutorial, I don't think we should hide away this contextual material on sequence data in a disclosable element. It seems crucial to understanding the data used by Nextstrain and the data requirements.

src/tutorials/quickstart.rst
```diff
@@ -1,34 +1,35 @@
-===============
-Running a build
-===============
+===========================
+Quickstart - Zika evolution
+===========================
```
tsibley, line 3: My prior thinking on this doc's place in our "learning ramp", and thus its title, is in 3a8ca8bb812ccba58c34131f8b1de2f079e806f6. I still suggest avoiding the "quickstart" moniker and keeping the title descriptive of what the tutorial shows how to do. Ok to include Zika in the title, but also I think its an incidental detail borne of convenience rather than an important choice and don't think it's necessary to call out. Good to align with the next doc in the ramp, however, as you've done.

So perhaps _Running a pathogen workflow_ and _Creating a pathogen workflow_?

To then align the TB-based tutorial's name, it could be _Creating a pathogen workflow (VCF)_ maybe?

tsibley (commented):
This is great! I've wanted to start a glossary for a while but haven't managed to so I'm glad you found the time and energy to sort through and rectify our (confusing) terms! :100:

My main comment is about titles for the Zika-based tutorials (see below).

huddlej:
I had the exact same feedback as @tsibley, but he said it better than I would have. I'm really excited about this update!

src/tutorials/quickstart.rst
jameshadfield: > So perhaps Running a pathogen workflow and Creating a pathogen workflow?

This is nice, and we should emphasise that these are _the same_ workflow, with the former deliberately ignoring the individual steps in the pipeline and the latter explaining each individual step.

> To then align the TB-based tutorial's name

In a previous meeting we decided to [Push TB down the (sidebar) order](https://docs.google.com/document/d/1HxnvQew-2y3ZIB4LWPlHle3nJGQiDTS0uP1uxoAtAuI/edit#), which I think could be done in this PR. I'd suggest a name of _Creating a bacterial pathogen workflow_. (That tutorial also needs some updating, but that's another issue!)

src/reference/glossary.rst
```diff
@@ -0,0 +1,75 @@
...
+   Auspice
+
+      A web application used for phylogenetic visualization and analysis. :doc:`Documentation<auspice:index>`
+
+   workflow
+      also *pathogen workflow*, *pathogen analysis*, *Nextstrain workflow*
+
+      A reproducible process comprised of one or more :term:`builds<build>` producing :term:`datasets<dataset>`, which can be visualized by :term:`Auspice`. Implementation varies per workflow, but generally they are run by workflow managers such as Snakemake.
+
+      Our core workflows can be divided into two types:
```
jameshadfield, line 20: Let's also define "core" 😄 https://docs.nextstrain.org/en/latest/guides/share/index.html is the relevant page, although the word "core" doesn't appear there at all!

src/reference/glossary.rst
```diff
@@ -0,0 +1,75 @@
...
+      :
+
+      - ``flu_seasonal_h3n2_ha_2y_meta.json``
+      - ``flu_seasonal_h3n2_ha_2y_tree.json``
+      - ``flu_seasonal_h3n2_ha_2y_tip-frequencies.json``
+
+      Some :term:`workflows<workflow>` produce a single, synonymous dataset, like Zika. Others, like seasonal flu, produce many datasets.
+
+   JSONs
+      Special ``.json`` files produced and consumed by :term:`Augur` and visualized by :term:`Auspice`. These files make up a :term:`dataset`.
```
jameshadfield, line 58: Minor point, but the (JSON) files visualised by Auspice are not the files consumed by Augur, although I'll note that some scripts may use them.

We have used the term "Auspice JSONs" to refer to the "dataset" in this glossary.

We also use the term "Node data JSONs" which are the JSONs consumed by augur (but not auspice).

src/reference/glossary.rst
```diff
@@ -0,0 +1,75 @@
...
+      also *Nextstrain CLI*
+
+      The Nextstrain command-line interface (**Nextstrain CLI**) provides a consistent way to run and visualize :term:`pathogen builds<Build>` and access Nextstrain components like :term:`Augur` and :term:`Auspice` across :term:`runtimes<runtime>` such as Docker, Native, and AWS Batch.
+
+      :doc:`Documentation <cli:index>`
+
+   runtime
+      also *Nextstrain runtime*
+
+      When installing and using Nextstrain, there are different configuration options, or **runtimes**, depending on the operating system.
```
jameshadfield, line 71: Can we change "Nextstrain" to "the Nextstrain CLI" (as described above), or use `nextstrain` to indicate it's the CLI not the catch-all term for the project.

src/tutorials/zika.rst
victorlin: The first time I went through this, this explanatory text was so long that I got lost where in the sequential process I was. I think the new table of contents improves that experience, so I've reverted the change here: 527d1efbd1d2dd57cae5f1f8cc386c3d6da49190

src/reference/glossary.rst
victorlin: Done: 87872171e9ecc1972c88e9e05e2aed167aa1e884

src/reference/glossary.rst
victorlin: Removed the reference to consumption by Augur: 8e509bc35bd0bc4f69f34db76e04bb36ddb77a70

Added "Auspice JSONs" as an alternate name to "dataset": 8275fe2aec16d630a57e1a9f8f23e6184d78cefd

I don't see "Node data JSONs" referenced anywhere in the tutorials, so I'll make a note to include this one later.

src/reference/glossary.rst
victorlin: Added "core workflow": e3099ca619da192d18f891241e901d7f748ce0ed

We can update docs to use the new term in another PR.

src/tutorials/quickstart.rst
tsibley: Yep, we should definitely setup redirects. The process for setting them up is [documented in this repo's readme](https://github.com/nextstrain/docs.nextstrain.org#configuring-redirects).

(Ideally we'd maintain redirects in a file in this repo and use the RTD v3 API to keep the live redirects automatically synced with the file, but that's not yet reality.)

tsibley (approved):
This looks great to me :tada:, but before merge/deploy, we'll want:

- [x] [Setup page redirects](https://github.com/nextstrain/docs.nextstrain.org/pull/84#discussion_r779779060) for renamed files
- [ ] Input from @jameshadfield on the [title of the TB-based tutorial](https://github.com/nextstrain/docs.nextstrain.org/pull/84#discussion_r779265782).

src/tutorials/quickstart.rst
jameshadfield: > Does bacterial pathogen workflow here simply mean VCF in place of FASTA, or is it more complicated than that?

The TB workflow is more of an introductory bacterial workflow, in the same way the zika tutorial is an introductory viral workflow. As well as VCF inputs, it includes inference of anti-microbial resistance.

tsibley:
> (Ideally we'd maintain redirects in a file in this repo and use the RTD v3 API to keep the live redirects automatically synced with the file, but that's not yet reality.)

I ended up implementing this sync… and just used it to successfully create these redirects. As the readme for this repo notes, they work now for the PR RTD build preview and don't interfere with the production docs, so I'll leave them there.

I'll push up the sync code momentarily in a separate repo. Later, we can automate it here with a GitHub Action (if we want).

src/tutorials/quickstart.rst
victorlin: @jameshadfield Ok, I assume your title will be more applicable to our readers. Updated and renamed file: 625640f2628b6b2386be2f03e97c2f258868235a

Also reordered sidebar as suggested: 2b9d03b895caca6171b12d0656b1165dc656c6e0

victorlin:
@tsibley thanks, that's great!

I have a messy commit history on this branch, so I was hoping to squash and merge. Since you added 6110a7865725fc851dfe8aa7afcfc35578f96503, would you be OK with that? Otherwise, I can try to rebase and clean it up a bit for a merge commit.

victorlin:
@jameshadfield if the newest changes look good to you, I'll merge this.

jameshadfield (approved):
https://nextstrain--84.org.readthedocs.build/en/84/ looks good to me, thanks Victor!

victorlin:
@tsibley the new redirects aren't working now. I'm not familiar with our RTD setup, but making a guess: the PR instances inherit redirects from the main project. So they won't work in PR instance but once merged, a new run of [`sync-redirects`](https://github.com/nextstrain/docs.nextstrain.org/blob/54dc3229489aa2a92ac80cd1d5401455f6c99a32/.github/workflows/sync-redirects.yaml) will add them. Is that correct?

---

> Note that it is safe to have redirects defined for pages that haven't yet moved, because RTD only consults page redirects when handling a 404 (our readme discusses this a little).

And with [this behavior](https://github.com/nextstrain/docs.nextstrain.org/pull/86#discussion_r780510208), I'm assuming the proper way to test these in the future is to manually (and selectively) run sync with the main project to check? Obviously, we don't want to automatically run sync with any PR, for example one with an empty redirects file that would adversely impact the main project instance.

tsibley:
@victorlin Yes, that's correct. The redirects for this PR only worked in the PR preview because I'd manually synced them. They were removed when we split out the new redirects from the redirect sync PR (#86) and merged it.

> I'm assuming the proper way to test these in the future is to manually (and selectively) run sync with the main project to check?

Yep! That's why the workflow is triggerable via the GitHub Actions UI (responds to the `workflow_dispatch` event).

tsibley:
RTD's model for redirects is that they exist at the project level, not per build/version of the project. ["Page" redirects](https://docs.readthedocs.io/en/stable/user-defined-redirects.html#page-redirects) apply to each build/version. ["Exact" redirects](https://docs.readthedocs.io/en/stable/user-defined-redirects.html#exact-redirects) apply to a single build/version, so we _might_ be able to automagically rescope new redirects in a PR to apply only to the PR build/version, e.g. a "Page" redirect from `/tutorials/quickstart.html` becomes an ["Exact" redirect](https://docs.readthedocs.io/en/stable/user-defined-redirects.html#exact-redirects) from `/en/84/tutorials/quickstart.html`.

victorlin:
@tsibley cool! I just [manually triggered](https://github.com/nextstrain/docs.nextstrain.org/actions/runs/1679770504) it on this PR branch and confirmed it works. Merging now.

PR #85, opened 2022-01-06
Auto-detect current year for copyright text

victorlin:
This change pulls the current year from [`datetime.now()`](https://docs.python.org/3/library/datetime.html#datetime.datetime.now) for the copyright text.

### DIFF
--- a/src/conf.py
+++ b/src/conf.py
@@ -21,8 +21,9 @@ subprocess.check_call([sys.executable or "python", './fetch-docs.py'])
 
 # -- Project information -----------------------------------------------------
 
+from datetime import datetime
 project = 'Nextstrain'
-copyright = '2021, Trevor Bedford and Richard Neher'
+copyright = f'{datetime.now().year}, Trevor Bedford and Richard Neher'
 author = 'The Nextstrain Team'
 
 


tsibley (approved):
:+1: for the little things. :-)

A note for anyone wondering (probably no one): my understanding from past legal advice is that templating the copyright year in this case is kosher because it's only updated when the docs are regenerated/republished. (It would not be kosher, for example, to dynamically display the current year each time the page is _viewed_.)

victorlin:
> A note for anyone wondering (probably no one): my understanding from past legal advice is that templating the copyright year in this case is kosher because it's only updated when the docs are regenerated/republished.

Thanks, I hadn't thought of that. Glad it worked out :)

> We also don't need to preserve all the years of first publication (or the range) in the copyright since we version the docs and past versions will retain their original copyright years.

Do we version in this repo? I don't see any tags and the only option on the page is `v: latest`.

Some follow-up remarks from a quick scan of other repos:

1. Auspice date is hardcoded: [auspice/docs/conf.py](https://github.com/nextstrain/auspice/blob/efe6fffc29f29f3ce353b796af6ba774fdb3bccb/docs/conf.py#L21) I can make the same change there, but want to understand the below first.
2. Augur is also using `datetime` to pull the year ([augur/docs/conf.py](https://github.com/nextstrain/augur/blame/73a07cfb39140a595d2a16ce1937a6bc54139db0/docs/conf.py#L47)). But how does this preserve the date to differ between version tags (e.g. [10.0.0](https://docs.nextstrain.org/projects/augur/en/10.0.0/) vs. [13.0.0](https://docs.nextstrain.org/projects/augur/en/13.0.0/)?) Does RTD generate the page files on tag push, then cache that forever? I'd assume any re-generation using `date.today().year` won't be smart enough to set the clock back to date of tag.

tsibley:
> Do we version in this repo? I don't see any tags and the only option on the page is `v: latest`.

Ahhh, right. I was thinking this was in a software repo that we version. So you're right that we don't here…

> 2. Augur is also using `datetime` to pull the year ([augur/docs/conf.py](https://github.com/nextstrain/augur/blame/73a07cfb39140a595d2a16ce1937a6bc54139db0/docs/conf.py#L47)). But how does this preserve the date to differ between version tags (e.g. [10.0.0](https://docs.nextstrain.org/projects/augur/en/10.0.0/) vs. [13.0.0](https://docs.nextstrain.org/projects/augur/en/13.0.0/)?) Does RTD generate the page files on tag push, then cache that forever? I'd assume any re-generation using `date.today().year` won't be smart enough to set the clock back to date of tag.

Yes, RTD only generates once for a tag and then stores ~forever. If we manually triggered regeneration of a past version, it would indeed bake in the current year. Not sure if this is kosher or not, but also not sure how much it really matters/how likely it is.

PR #86, opened 2022-01-07
Manage RTD redirects using a file in this repo

tsibley:
Automatically synced to RTD using the new readthedocs-cli package I wrote and a new GitHub Actions workflow for this repo.

Includes the `redirects.yml` file I pushed to #84, so it doesn't matter which order merge happens in.

I setup the `RTD_TOKEN` secret on GitHub (using the `nextstrain` RTD user's account) and manually triggered [a test run of the workflow](https://github.com/nextstrain/docs.nextstrain.org/runs/4743056090?check_suite_focus=true). All looks well.

### DIFF
--- a/redirects.yml
+++ b/redirects.yml
@@ -1,0 +1,156 @@
+# Authoritative list of redirects we have configured in RTD.  See this repo's
+# README.md¹ for more information on maintaining redirects.
+#
+# ¹ https://github.com/nextstrain/docs.nextstrain.org#configuring-redirects
+---
+- type: page
+  from_url: /reference/formats/
+  to_url: /reference/data-formats.html
+
+- type: page
+  from_url: /reference/
+  to_url: /
+
+- type: page
+  from_url: /guides/
+  to_url: /
+
+- type: page
+  from_url: /tutorials/
+  to_url: /
+
+- type: page
+  from_url: /learn/
+  to_url: /
+
+- type: page
+  from_url: /reference/index.html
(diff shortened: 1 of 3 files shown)


redirects.yml
```diff
@@ -0,0 +1,168 @@
...
+  from_url: /tutorials/quickstart.html
+  to_url: /tutorials/running-a-workflow.html
+
+- type: page
+  from_url: /tutorials/zika.html
+  to_url: /tutorials/creating-a-workflow.html
+
+- type: page
+  from_url: /tutorials/tb_tutorial.html
+  to_url: /tutorials/creating-a-bacterial-pathogen-workflow.html
```
victorlin, line 168: If you remove these, we can merge this PR before #84, then I can add them in there.

victorlin (changes requested):

redirects.yml
tsibley: Sure.

Note that it is safe to have redirects defined for pages that haven't yet moved, because RTD only consult page redirects when handling a 404 (our readme [discusses this](https://github.com/nextstrain/docs.nextstrain.org#configuring-redirects) a little).

tsibley: Done.

victorlin (approved):

redirects.yml
victorlin: @tsibley oh cool, well this will make for better git commit history at the least. Thanks!

PR #87, opened 2022-01-14
glossary: Add "narrative" term

tsibley:


src/reference/glossary.rst
```diff
@@ -60,6 +60,12 @@ Glossary
 
       Some :term:`workflows<workflow>` produce a single, synonymous dataset, like Zika. Others, like seasonal flu, produce many datasets.
 
+   narrative
+
+      A Markdown file with additional syntax to specify a :term:`dataset` to display alongside the text.
```
victorlin, line 65: This is a perfect technical definition, but maybe not so relevant to new users?

Suggest adding something to describe the multi-page functionality of narratives (yes it's a single file, but is more of a slideshow to readers). Example:

> A method to convey information through multiple pages of content. Datasets can be displayed alongside text as interactive Auspice views. Saved as a Markdown file with extended syntax to support additional displays.

src/reference/glossary.rst
```diff
@@ -60,6 +60,12 @@ Glossary
...
+
+      Viewable on nextstrain.org or with :term:`Auspice` via the :doc:`cli:commands/view` or :doc:`auspice view <auspice:introduction/how-to-run>` commands.
+
```
victorlin, line 68: Link out to existing docs?

- https://docs.nextstrain.org/en/latest/guides/communicate/narratives-intro.html
- https://docs.nextstrain.org/en/latest/tutorials/narratives-how-to-write.html

victorlin (changes requested):

src/reference/glossary.rst
tsibley: Fair! Thanks for the nudge. Revised and repushed.

jameshadfield (approved):

victorlin (approved):

PR #89, opened 2022-03-03
Document changes to group-overview.md frontmatter

tsibley:
As of <https://github.com/nextstrain/nextstrain.org/pull/478>.  Thanks
to @joverlee521 for the prompt to update this.


joverlee521 (approved):
Thanks for making the change!

PR #98, opened 2022-03-16
Update conda environment

### LINKED ISSUE #97: Lists not rendering
Same thing as nextstrain/ncov#891 but for this RTD project. See [Installing Nextstrain](https://docs.nextstrain.org/en/latest/install.html) as an example of a page that is improperly rendered.

victorlin:
_same thing as nextstrain/ncov#890 for this RTD project._

Fixes #97.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -7,10 +7,10 @@ dependencies:
   - recommonmark
   - requests
   - sphinx
+  - sphinx-markdown-tables
+  - sphinx-argparse
   - sphinx-tabs
   - pip
   - docutils<0.17 # https://github.com/nextstrain/ncov/issues/891
   - pip:
     - nextstrain-sphinx-theme>=2020.2
-    - sphinx-markdown-tables
-    - sphinx-argparse
--- a/environment.yml
+++ b/environment.yml
@@ -1,15 +1,16 @@
 name: docs.nextstrain.org
 channels:
-  - defaults
+  - conda-forge
 dependencies:
   - graphviz
   - make
   - recommonmark
   - requests
   - sphinx
+  - sphinx-tabs
   - pip
+  - docutils<0.17 # https://github.com/nextstrain/ncov/issues/891
   - pip:
     - nextstrain-sphinx-theme>=2020.2
     - sphinx-markdown-tables
     - sphinx-argparse
-    - sphinx-tabs


tsibley (approved):
Looks ok to me, but would be nice to include a summary of _why_ the switch from pip → conda was made in the actual commit message of 1700211.

tsibley:
Between comments on issue https://github.com/nextstrain/ncov/pull/891, comments on PR https://github.com/nextstrain/ncov/pull/890, and the commit messages in both that PR and this one, I still don't understand if the switching to Conda packages it critical somehow. Like, would this change by itself be sufficient:

```diff
diff --git a/environment.yml b/environment.yml
index fafa435..3b5ef43 100644
--- a/environment.yml
+++ b/environment.yml
@@ -11,5 +11,6 @@ dependencies:
   - pip:
     - nextstrain-sphinx-theme>=2020.2
     - sphinx-markdown-tables
     - sphinx-argparse
     - sphinx-tabs
+    - docutils<0.17 # https://github.com/nextstrain/ncov/issues/891
```

or not?

It's fine to switch to Conda packages, I just can't tell if it's a load bearing change or a policy change (e.g. prefer Conda over Pip when using Conda).

victorlin:
@tsibley good point, I've amended the commits to have more clear reasoning.

However, after looking into the `docutils=0.17` issue further, there may be a better solution to all this. Will not merge this yet.

victorlin:
Closing since the issue can be fixed in a better way: https://github.com/nextstrain/ncov/issues/891#issuecomment-1069903910

tsibley:
> Closing since the issue can be fixed in a better way

:+1: 

Thanks for the explanations anyway!

> This is only possible if sphinx-tabs is installed with conda, otherwise pip would resolve the latest sphinx-tabs which requires docutils=0.17 as of version 3.3.0 [2] and re-install docutils=0.17 over any existing version (third example here [3]).

FWIW, this is dependent on the version of `pip`, specifically its dep resolver.

I get the right behaviour with the [new dep resolver in pip >=20.3]([https://pip.pypa.io/en/stable/user_guide/#changes-to-the-pip-dependency-resolver-in-20-3-2020):

```
$ pip --version
pip 21.3.1 from /tmp/tmp.oEjPFgxX9g/lib/python3.6/site-packages/pip (python 3.6)
$ pip install sphinx-tabs docutils'<0.17'
…
Successfully installed … docutils-0.16 … sphinx-tabs-3.2.0 …
```

but this is exactly the case that the old resolver fails to handle properly:

```
$ pip --version
pip 20.3 from /tmp/tmp.z7vJnjMTgO/lib/python3.6/site-packages/pip (python 3.6)
$ pip install sphinx-tabs docutils'<0.17'
…
Successfully installed … docutils-0.17 … sphinx-tabs-3.3.0 …
```

It does notice the problem when I re-run the same `pip install` again:

```
$ pip install sphinx-tabs docutils'<0.17'
…
Installing collected packages: docutils
  Attempting uninstall: docutils
    Found existing installation: docutils 0.17
    Uninstalling docutils-0.17:
      Successfully uninstalled docutils-0.17
ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.

We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.

sphinx-tabs 3.3.0 requires docutils~=0.17.0, but you'll have docutils 0.16 which is incompatible.
```

PR #99, opened 2022-03-18
Warn about slow macOS+Docker on install page

victorlin:
More info: https://github.com/nextstrain/docker-base/issues/35

### DIFF
--- a/src/install.rst
+++ b/src/install.rst
@@ -70,6 +70,12 @@ These instructions will install the Nextstrain CLI and tools to run and view you
       5. Install the remaining Nextstrain components. There are two options:
 
          a. Docker (recommended) – install Docker Desktop using `the official guide <https://docs.docker.com/desktop/mac/install/>`_.
+
+            .. warning::
+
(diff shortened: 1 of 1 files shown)


src/install.rst
```diff
@@ -69,7 +69,13 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
       5. Install the remaining Nextstrain components. There are two options:
 
-         a. Docker (recommended) – install Docker Desktop using `the official guide <https://docs.docker.com/desktop/mac/install/>`_.
+         a. Docker (recommended :superscript:`\*` ) – install Docker Desktop using `the official guide <https://docs.docker.com/desktop/mac/install/>`_.
```
tsibley, line 72: I think the superscript asterisk is unnecessary here since the warning is bright orange and immediately below this line. (If we keep the asterisk, let's remove the superscript: it makes rendering worse.)

Alternatively, to reduce clutter, we could switch to a footnote and put the warning there.

tsibley (changes requested):

src/install.rst
victorlin: Removed superscript asterisk. I'd say good to keep the warning closer to the impacted step since share of M1 users will only increase with time.

tsibley: Sounds good. I would hope that it's not much longer before we address nextstrain/docker-base#35!

tsibley (approved):

Issue #100, opened 2022-03-18
Standardize and clarify terminology


victorlin:
Inspired by discussion at https://github.com/nextstrain/docs.nextstrain.org/pull/92#discussion_r826237041, this issue tracks efforts to standardize and clarify terminology across documentation.

Progress should be reflected by modifications to [Glossary](https://docs.nextstrain.org/en/latest/reference/glossary.html) [[src](https://github.com/nextstrain/docs.nextstrain.org/blob/master/src/reference/glossary.rst?plain=1)].

victorlin:
Having a glossary is a good start. Closing this since the issue is not clearly scoped. We can consider re-opening under a more specific title like "Ensure every existing glossary entry is properly used across all pages in all docs projects".

PR #101, opened 2022-03-18
Replace build term

victorlin:
[_docs preview_](https://nextstrain--101.org.readthedocs.build/en/101/)

This PR replaces the **_build_** term with either **_workflow_** or **_dataset_**, depending on context.

### Related issue(s)

- #100

### Notes

To truly free up **_build_** from our terminology, the following should change but I'm not sure if we want to pursue:

- _community builds_ → _community datasets_
- `nextstrain build` → `nextstrain run` (?)
- ~ncov: `builds.yaml` → `config.yaml`, `builds:` → `outputs:`~ see https://github.com/nextstrain/docs.nextstrain.org/pull/92#discussion_r826237041

I noticed [local-vs-container.svg](https://github.com/nextstrain/docs.nextstrain.org/blob/a3e57c00e8d04d903e9cfe17a34ae913937bd1c2/src/images/local-vs-container.svg) has references to **_build_**, but I can't find any references to this file. Is it being used?


tsibley:
Thanks for drafting this up, @victorlin! This is a change with broader implications, so we'll want to ([continue to!](https://github.com/nextstrain/docs.nextstrain.org/pull/92#discussion_r826237041)) gather broad input from the team on it and consider if it's even a switch we want to make. I think we want something approaching consensus for this before merge rather than the typical "1 approval and merge" process we typically use.

victorlin:
@tsibley for sure! Until we reach consensus, this PR serves to preview how things could look if we decide to go down this path.

victorlin:
Closing since I don't think there's much need to push these changes through.

PR #102, opened 2022-03-21
Update docker/native installation steps

victorlin:
_[preview](https://nextstrain--102.org.readthedocs.build/en/102/install.html)_

1. Use separate `mamba create` for docker vs. native
    - This helps resolve dependencies properly for native, and avoids Python pinning issues such as [1] where Python 3.10 is installed+pinned as latest available with nextstrain-cli, but one of augur's dependencies does not yet support 3.10. Creating the environment with all packages listed will properly resolve to 3.9.
2. Separate install tabs for Docker vs. native.

[1]: https://github.com/bioconda/bioconda-recipes/pull/33757#issuecomment-1072675574.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -13,3 +13,4 @@ dependencies:
     - sphinx-markdown-tables
     - sphinx-argparse
     - sphinx-tabs
+    - sphinx-copybutton
--- a/src/conf.py
+++ b/src/conf.py
@@ -40,6 +40,7 @@ extensions = [
     'sphinx.ext.autodoc',
     'sphinx_tabs.tabs',
     'sphinx.ext.graphviz',
+    'sphinx_copybutton',
 ]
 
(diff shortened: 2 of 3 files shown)


huddlej (commented):
This looks really good, @victorlin. The tabs present the forking path much more clearly, that this change would be worthwhile even without the Conda technicality motivating it.

A minor concern is that the "Native" Conda commands extend beyond the right margin of the code box, so the user has to scroll right to see there is more to copy/paste.

image

Could we add line returns to the commands to show the whole command in a single view like:

```
mamba create -n nextstrain \
  -c bioconda nextstrain-cli \
  --yes
  augur auspice nextalign snakemake git
```

That could be a separate PR, if you prefer.

victorlin:
@huddlej thanks for the feedback! I added line breaks along with a copy button, which should render a better experience there.

j23414 (approved):

PR #104, opened 2022-04-05
Overhaul Nextstrain Groups documentation

tsibley:
See commit messages for details. Motivated by upcoming Groups migration. There's more to improve here, for sure, but I wanted to pick a reasonable stopping point and get what I had out for review.

Diffs in some individual commits are useful to read, but you'll definitely want to read through the end result in [the preview build](https://nextstrain--104.org.readthedocs.build/en/104/index.html) too.

Pages with significant changes:

- [learn/groups/index](https://nextstrain--104.org.readthedocs.build/en/104/learn/groups/index.html)
- [guides/share/index](https://nextstrain--104.org.readthedocs.build/en/104/guides/share/index.html)
- [guides/share/groups/index](https://nextstrain--104.org.readthedocs.build/en/104/guides/share/groups/index.html)
- [guides/share/groups/customize](https://nextstrain--104.org.readthedocs.build/en/104/guides/share/groups/customize.html)
- [guides/share/groups/migrate-from-s3](https://nextstrain--104.org.readthedocs.build/en/104/guides/share/groups/migrate-from-s3.html)

### Related issue(s)
https://github.com/nextstrain/private/issues/20


src/guides/share/groups/customize.rst
```diff
@@ -0,0 +1,63 @@
+.. _group-customization:
+
+===========================
+Customize your group's page
+===========================
+
+.. hint::
```
jameshadfield, line 7: We should mention that the user has to have owner permissions, as per

> Owners ... same permissions as editors plus the ability to manage the group's description and logo

Or relax this requirement -- I think editors should be able to edit the group's customisations.

jameshadfield (commented):
Thanks for the docs Tom. I think they describe well the current CLI behavior. Reviewing this gave me the opportunity to finally test out the CLI behavior and there are some things I think we can improve there. I'll write them up as a CLI issue (shortly) rather than overloading this PR.

joverlee521 (approved):
Thanks for writing this up! We should make note to internal members to honor the group "Owners" permissions when answering groups email requests. Not sure where is the best place to put that reminder...

tsibley:
> We should make note to internal members to honor the group "Owners" permissions when answering groups email requests. Not sure where is the best place to put that reminder...

Yes, definitely. I'd been pondering this too and related SOPs, and I think the best place is for policies like this is our wiki.

src/guides/share/groups/customize.rst
tsibley: Ah, good point. Added a mention in 3702799.

I don't agree about editors (at least currently) but am open to discussing roles more, esp. with the broader team. The current permissions gap between editors and owners is intentionally big so that there's room to add new roles between/adjacent to them in the future without breaking people's existing memberships. e.g. imagine a `managers` role that gets `editors` + customizing + membership management but not deletion of the group, or something similar. or a `curators` role that gets customization-only. Tightly-scoped roles now mean we can more easily change later. (and roles don't even have to be mutually exclusive, possibly shouldn't be, which is served better by tightly-scoped roles.)

tsibley:
@jameshadfield @joverlee521 Thoughts on merging this now vs. closer to when we actually send out migration notifications? I'm inclined to merge now so the material is available for the groups we migrate early (and the other unrelated docs improvements).

tsibley:
@joverlee521 Added [guidelines for handling requests](https://wiki.nextstrain.org/t/request-guidelines) to our wiki.

jameshadfield:
> Thoughts on merging this now vs. closer to when we actually send out migration notifications?

I'm happy to merge these docs as-is, as they describe the current behavior. [I would like us to change that behavior, however](https://github.com/nextstrain/cli/issues/169), and this should be done before we migrate any non-internal groups.

tsibley:
@jameshadfield Ok, I'll merge this now and reply fully to your review of `nextstrain remote` behaviour after I'm back.

PR #105, opened 2022-04-06
Add instructions on how to build subproject docs

jameshadfield:
While these docs belong in each subproject (repo), it's nice
to provide pointers here.

~I couldn't find documentation on how to build the CLI docs locally~ There is no documentation for building the CLI docs locally, see https://github.com/nextstrain/cli/issues/131.

### DIFF
--- a/README.md
+++ b/README.md
@@ -36,6 +36,8 @@ You can view them by running:
 Leave the environment with:
 
     conda deactivate
+
+> See below for how to build the docs for each subproject
     
 ## Building the docs with Docker
 
@@ -145,6 +147,18 @@ Documents which we've kept in subprojects, as opposed to including them in this 
 Another example is the tutorial for the SARS-CoV-2 workflow, which lives alongside the workflow in its repository, https://github.com/nextstrain/ncov, and shows up as a subproject at https://docs.nextstrain.org/projects/ncov/en/latest/index.html.
 Subprojects necessitate a distinct URL, such as https://docs.nextstrain.org/projects/augur as opposed to just https://docs.nextstrain.org.
 By default, this means a distinct set of headings / links in the sidebar navigation, making navigating back to https://docs.nextstrain.org more difficult once you have navigated to a subproject.
+
+## Building the Subproject docs
+
+The documentation for each subproject is built from within the corresponding GitHub repo, also using sphinx.
+
+For full details, see the respective how-to-build-docs sections:
+
+- [Augur](https://github.com/nextstrain/augur/blob/master/docs/contribute/DEV_DOCS.md#building-documentation)
+- [Auspice](https://github.com/nextstrain/auspice/blob/master/DEV_DOCS.md#contributing-to-documentation)
+- [Nextclade (includes Nextclade Web, Nextclade CLI, Nextalign CLI)](https://github.com/nextstrain/nextclade/tree/master/docs#building-the-docs-locally)
+- [ncov (SARS-CoV-2 Workflow)](https://github.com/nextstrain/ncov/tree/master/docs#building-the-docs)
+
 
 ### Fetching of documents from other repositories
 


README.md
```diff
@@ -146,6 +148,23 @@ Another example is the tutorial for the SARS-CoV-2 workflow, which lives alongsi
...
 
+## Building the SubProject docs
+
+Typically the documentation for each subproject is build using the same commands as the umbrella docs described above:
+```
+cd <repo>/docs
+make clean # optional
+make html
+```
+This may work in the `docs.nextstrain.org` conda environment, but some subprojects require additional dependencies.
```
tsibley, line 159: Hmm, I don't think we should suggest using the docs.nextstrain.org Conda env, even with that caveat. It's never the correct thing to do, even if it appears to work.

README.md
```diff
@@ -146,6 +148,23 @@ Another example is the tutorial for the SARS-CoV-2 workflow, which lives alongsi
 Subprojects necessitate a distinct URL, such as https://docs.nextstrain.org/projects/augur as opposed to just https://docs.nextstrain.org.
 By default, this means a distinct set of headings / links in the sidebar navigation, making navigating back to https://docs.nextstrain.org more difficult once you have navigated to a subproject.
 
+## Building the SubProject docs
```
tsibley, line 151:
```suggestion
## Building the Subproject docs
```

README.md
```diff
@@ -146,6 +148,23 @@ Another example is the tutorial for the SARS-CoV-2 workflow, which lives alongsi
...
+
+Typically the documentation for each subproject is build using the same commands as the umbrella docs described above:
+```
+cd <repo>/docs
+make clean # optional
+make html
+```
```
tsibley, line 158: I would omit these pseudo-instructions since they're not specific to a project and just keep the (below) links out to each subproject's own how-to for building its docs.

README.md
```diff
@@ -146,6 +148,23 @@ Another example is the tutorial for the SARS-CoV-2 workflow, which lives alongsi
...
+This may work in the `docs.nextstrain.org` conda environment, but some subprojects require additional dependencies.
+For full details, see the respective how-to-build-docs sections:
+
+- [Augur](https://github.com/nextstrain/augur/blob/master/docs/contribute/DEV_DOCS.md#building-documentation)
+- [Auspice](https://github.com/nextstrain/auspice/blob/master/DEV_DOCS.md#contributing-to-documentation)
+- [Nextclade (includes Nextclade Web, Nextclade CLI, Nextalign CLI)](https://github.com/nextstrain/nextclade/tree/master/docs#building-the-docs-locally)
+- [ncov (SARS-CoV-2 Workflow)](https://github.com/nextstrain/ncov/tree/master/docs#building-the-docs)
```
tsibley, line 165: Not something to address with this PR, but seeing these links to GitHub instead of https://docs.nextstrain.org makes me think we'd be better off in the long run making our dev docs (including this README here) part of https://docs.nextstrain.org and using cross-refs. Helps make the links more robust to breaking and makes it clear if they do break, while also surfacing more of our dev docs.

tsibley (changes requested):

README.md
jameshadfield: Resolved in 77f7f74

README.md
jameshadfield: Resolved in 77f7f74

README.md
jameshadfield: Resolved in 77f7f74

jameshadfield: It would be nice to have consistency (e.g. all directories named `docs`, all using `make html`) but I don't have the appetite to go changing that in a bunch of repos right now!

README.md
jameshadfield: 👍

PR #106, opened 2022-04-06
Clarify recommendations for installation options

victorlin:
_[preview](https://nextstrain--106.org.readthedocs.build/en/106/install.html)_

Previously, the text:

1. Recommended Docker without giving any explanation
2. Did not direct Windows users how to choose between Windows and WSL on Windows

This change:

1. Adds a note on how to choose between Docker vs. Native and Windows vs. WSL on Windows
2. Removes the universal recommendation for Docker

### Related issue(s)

_N/A_

### Testing

_N/A_

### DIFF
--- a/src/install.rst
+++ b/src/install.rst
@@ -36,7 +36,19 @@ When you use Docker to run Nextstrain components, you don’t need to manage any
 Installation Steps
 ==================
 
-These instructions will install the Nextstrain CLI and tools to run and view your own Nextstrain analyses. Configuration options vary by operating system.
+These instructions will install the Nextstrain CLI and tools to run and view your own Nextstrain analyses. Configuration options vary by platform.
+
+.. note::
+
+   For each operating system, there are two options:
+
+   1. **Docker**: use if you plan to use Nextstrain without other bioinformatics tools. Docker provides the most predictable environment.
+   2. **Native**: use if you plan to use other bioinformatics tools alongside Nextstrain. Native installation is done via conda, allowing flexibility in managing your own conda environments.
+
+   For Windows users, you may pick between:
+
+   1. **Windows** direct install: use if you choose **Docker** above, or prefer a PowerShell interface.
+   2. **WSL on Windows**: use if you choose **Native** above, or prefer a UNIX-like interface.
 
 .. tabs::
 
@@ -64,7 +76,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
          .. tabs::
 
-            .. group-tab:: Docker (recommended)
+            .. group-tab:: Docker
 
                .. warning::
 
@@ -186,7 +198,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
          .. tabs::
 
-            .. group-tab:: Docker (recommended)
+            .. group-tab:: Docker
 
                1. On Windows, `install Docker Desktop for WSL 2 backend <https://docs.docker.com/desktop/windows/wsl/>`_.
 
@@ -260,7 +272,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
          .. tabs::
 
-            .. group-tab:: Docker (recommended)
+            .. group-tab:: Docker
 
                1. Install Docker Engine for Ubuntu using the `convenience script <https://docs.docker.com/engine/install/ubuntu/#install-using-the-convenience-script>`_:
 


src/install.rst
```diff
@@ -36,7 +36,19 @@ When you use Docker to run Nextstrain components, you don’t need to manage any
...
 
-These instructions will install the Nextstrain CLI and tools to run and view your own Nextstrain analyses. Configuration options vary by operating system.
+These instructions will install the Nextstrain CLI and tools to run and view your own Nextstrain analyses. Configuration options vary by platform.
+
+.. note::
+
+   For each operating system, there are two options:
+
+   1. **Docker**: use if you plan to use Nextstrain without other bioinformatics tools. Docker provides the most predictable environment.
+   2. **Native**: use if you plan to use other bioinformatics tools alongside Nextstrain. Native installation is done via conda, allowing flexibility in managing your own conda environments.
```
tsibley, line 46: Hmm. I think I know what you intend here, but as it's worded, this reads as potentially misleading to me. I think, but correct me if wrong, that the intent is to communicate that if you want to use other arbitrary bioinformatics tools _within a `nextstrain build` run_ then it'll be (much) easier to do that with "native" installation via Conda. That emphasized bit of scoping seems important to me, as there's nothing about either installation method that makes it harder or easier to use other bioinformatics tools on Nextstrain inputs before hand, outputs afterwards, etc.

It might also be worth mentioning in this note that _both_ options can be followed, if you know up-front you want the flexibility to switch between then or (more realistically perhaps) you realize later you initially choose the wrong option for your needs.

Relatedly, ISTM that we're going to say "use Conda if you want to use other bioinformatics tools in your `nextstrain build` run", then we should also probably include some information about how to actually do that (i.e. how to install other arbitrary things into the Conda env). This implies to me that a separate doc page might be in order, something like a how-to _Use Nextstrain with other bioinformatics tools_.

src/install.rst
```diff
@@ -36,7 +36,19 @@ When you use Docker to run Nextstrain components, you don’t need to manage any
...
+
+   For Windows users, you may pick between:
+
+   1. **Windows** direct install: use if you choose **Docker** above, or prefer a PowerShell interface.
+   2. **WSL on Windows**: use if you choose **Native** above, or prefer a UNIX-like interface.
```
j23414, line 51: I like the explanation on "when to choose WSL", based on my past experience helping researchers to get set up on a UNIX-like environment to use the HPC, and learn various bioinformatic toolsets.

j23414 (approved):

src/install.rst
huddlej: I agree with @tsibley about the potential for confusion. Some of this choice comes down to whether users want to use Docker/Conda or not. We've heard from folks at office hours who do not want to use Docker (or Conda), and they will choose the installation platform that they do prefer. For people who don't know what they prefer, it still makes sense to me to recommend Docker (and flag Docker as the recommended approach).

Other components of this choice that seem most relevant are well-described here:

  - "Docker provides the most predictable environment"
  -  managing your own conda environments

In the past we've talked about this kind of decision as "if you know what you're doing, here are your options". The same context applies to the Windows vs. Windows/WSL choice.

We should discuss this more as a small group and refine this PR based on our discussion...

huddlej: I agree with @tsibley about the potential for confusion. Some of this choice comes down to whether users want to use Docker/Conda or not. We've heard from folks at office hours who do not want to use Docker (or Conda), and they will choose the installation platform that they do prefer. For people who don't know what they prefer, it still makes sense to me to recommend Docker (and flag Docker as the recommended approach).

Other components of this choice that seem most relevant are well-described here:

  - "Docker provides the most predictable environment"
  -  managing your own conda environments

In the past we've talked about this kind of decision as "if you know what you're doing, here are your options". The same context applies to the Windows vs. Windows/WSL choice.

We should discuss this more as a small group and refine this PR based on our discussion...

tsibley: > We should discuss this more as a small group and refine this PR based on our discussion...

+1

tsibley:
Bumping this.

j23414:
Is there anything blocking this from being merged? The text looks correct to me.

tsibley:
@j23414 There are unaddressed review comments above about what information to convey and how to convey it. I bumped this PR earlier only to bring attention to it again, not to suggest merge as-is. I do still think we want to have a small focused discussion on the topics brought up in the comments.

victorlin:
From [today's meeting](https://docs.google.com/document/d/1dODYr-3T8LAQt-ewMZKNrcy1Wvy70Fw-h8exXbIwbyM/), we plan to incorporate these recommendations into #123 and refrain from making any large changes to the installation docs outside of that PR, such as reordering the tab hierarchy. It is simplest to keep the existing recommendations until that PR is merged.

src/install.rst
tsibley: I'm using language like the above in https://github.com/nextstrain/docs.nextstrain.org/pull/123.

PR #111, opened 2022-05-24
Rework contributing pages

tsibley:
See commit messages for details of the two changes and the [revised Contributing page](https://nextstrain--111.org.readthedocs.build/en/111/guides/contribute/index.html) in the preview build.

### Testing
- [x] Builds locally and views ok
- [x] CI passes
- [x] Redirects work as expected

### DIFF
--- a/redirects.yml
+++ b/redirects.yml
@@ -182,3 +182,7 @@
 - type: page
   from_url: /guides/contribute/cli-develop.html
   to_url: https://github.com/nextstrain/cli/tree/master/doc/development.md#readme
+
+- type: page
+  from_url: /guides/contribute/documentation.html
+  to_url: /guides/contribute/index.html
--- a/redirects.yml
+++ b/redirects.yml
@@ -170,3 +170,15 @@
 - type: page
   from_url: /guides/share/nextstrain-groups.html
   to_url: /guides/share/groups/index.html
+
+- type: page
+  from_url: /guides/contribute/augur_develop.html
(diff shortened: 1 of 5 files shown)


tsibley:
Realized that since page redirects on RTD only apply if the page is missing, I could sync these purely additive redirects from this PR without affecting the stable version of the docs. I [ran the sync workflow](https://github.com/nextstrain/docs.nextstrain.org/runs/6579686747?check_suite_focus=true) and now this lets the redirects be tested/confirmed in the preview build:


- https://nextstrain--111.org.readthedocs.build/en/111/guides/contribute/augur_develop.html
- https://nextstrain--111.org.readthedocs.build/en/111/guides/contribute/auspice-develop.html
- https://nextstrain--111.org.readthedocs.build/en/111/guides/contribute/cli-develop.html
- https://nextstrain--111.org.readthedocs.build/en/111/guides/contribute/documentation.html

huddlej (approved):
> The best context for these documents is the repository they live in, not
the documentation website.  The documents are written (reasonably so!)
to be viewed/used alongside the code itself.  Linking to the docs on
GitHub helps facilitate this context and code browsing.

> This also reduces the number of documents we're transcluding, which I
see as a bonus since long-term I think we want to stop this practice (or
alternatively, double down on it and start doing it much more
systematically (e.g. with cross-project link resolution/rewriting)).

I agree strongly with both of these points. For some reason, I had thought we had discussed moving toward including our dev docs directly into docs.nextstrain.org, but linking to the GitHub context feels right here (and I can't find that previous discussion anywhere, so maybe I'm imagining it).

PR #115, opened 2022-06-23
Install SARS-CoV-2 workflow dependencies in Native runtime

### LINKED ISSUE #96: Streamline native installation steps for subproject references
### Context

In the ongoing installation documentation revolution, the initial thought was to de-duplicate installation instructions and have subprojects reference the [central installation guide](https://docs.nextstrain.org/en/latest/install.html).

This has already been done with ncov: [Setup and installation](https://docs.nextstrain.org/projects/ncov/en/latest/analysis/setup.html). However, the new page leads to confusion such as #93. Users are redirected to the central installation docs and follow all the steps there, thinking they are done with installation once finished with that page. They are likely to miss the additional note for native runtime:

ncov-installation-steps

### Possible solutions

1. Avoid extra installation steps
    - This is how Docker runtime is handled – it already installs ncov-specific tools that are the reason for additional installation with native runtime.
2. Copy the same installation guide into the project installation page. This would go back to the old model with duplication issues such as getting out of sync.
3. For Snakemake workflows, use the `--use-conda` flag. This would point to the conda environment YAML checked into the workflow-specific repos (e.g. [nextstrain.yaml in ncov](https://github.com/nextstrain/ncov/blob/master/workflow/envs/nextstrain.yaml)).

victorlin:
[_preview_](https://nextstrain--115.org.readthedocs.build/en/115/install.html)

The Nextstrain Docker image used by the Docker runtime comes with many
other software tools, so that the runtime can work with all core
pathogen workflows. In contrast, the current Native runtime installs a
"bare minimum" set of software, requiring individual workflows to
specify additional installation commands just for the Native runtime.
This commit effectively syncs the two runtimes to provide the same
tools.

In the long term, it would be best to abandon this approach of
supporting *n* pathogen workflows by a single setup (for both Docker and
Native runtime). While this might seem like taking a step backwards by
putting more potentially unused tools in the Native runtime, the
difference of available tools between Docker/Native currently causes
confusion for users and unnecessary instructions specific to the Native
runtime for each pathogen workflow.

### Related issue(s)

- Fixes #96
- Closes #116

### Testing

_N/A_

### TODO

- [x] create sibling PR in ncov repo to remove note on installing additional tools for Native runtime: nextstrain/ncov#967

### DIFF
--- a/src/install.rst
+++ b/src/install.rst
@@ -106,7 +106,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
                      mamba create -n nextstrain \
                        -c conda-forge -c bioconda \
-                       nextstrain-cli augur auspice nextalign snakemake git \
+                       nextstrain-cli augur auspice nextalign nextclade snakemake git epiweeks pangolin pangolearn \
                        --yes
 
                2. Activate the conda environment:
@@ -227,7 +227,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
                      mamba create -n nextstrain \
                        -c conda-forge -c bioconda \
-                       nextstrain-cli augur auspice nextalign snakemake git \
+                       nextstrain-cli augur auspice nextalign nextclade snakemake git epiweeks pangolin pangolearn \
                        --yes
 
                2. Activate the conda environment:
@@ -313,7 +313,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
                      mamba create -n nextstrain \
                        -c conda-forge -c bioconda \
-                       nextstrain-cli augur auspice nextalign snakemake git \
+                       nextstrain-cli augur auspice nextalign nextclade snakemake git epiweeks pangolin pangolearn \
                        --yes
 
                2. Activate the conda environment:


src/install.rst
```diff
@@ -106,7 +106,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
                      mamba create -n nextstrain \
                        -c conda-forge -c bioconda \
-                       nextstrain-cli augur auspice nextalign snakemake git \
+                       nextstrain-cli augur auspice nextalign snakemake git epiweeks pangolin pangolearn \
```
tsibley, line 109: _non-blocking_

Did you look into if adding these three packages significantly increases the installation time or size on disk?

tsibley (approved):
I'm fully on board with this short-term step and the long-term ideal described. :raised_hands:  I think it's better for users than more cross-linking (like #115 adds) but would still say that merge of this should wait until we have broadly concurring feedback from more of the team.

src/install.rst
victorlin: I hadn't, but that would be good to know! Here's some benchmarks:

### Installation time

Note that each `time` is taken after cleaning all caches so these are worst-case scenarios given my processor and network speed. Not much difference here (93.63s vs. 98.21s user+sys time).

```sh
conda clean --all --force-pkgs-dirs
time mamba create -n tmp1 \
    -c conda-forge -c bioconda \
    nextstrain-cli augur auspice nextalign snakemake git \
    --yes
# 52.33s user 41.30s system 92% cpu 1:41.63 total

conda clean --all --force-pkgs-dirs
time mamba create -n tmp2 \
    -c conda-forge -c bioconda \
    nextstrain-cli augur auspice nextalign snakemake git epiweeks pangolin pangolearn \
    --yes
# 55.48s user 42.73s system 90% cpu 1:48.14 total
```

### Disk usage

1.5G -> 1.6G (~0.1G increase).

```sh
du -hs /opt/homebrew/Caskroom/miniconda/base/envs/tmp1
# 1.5G	/opt/homebrew/Caskroom/miniconda/base/envs/tmp1
du -hs /opt/homebrew/Caskroom/miniconda/base/envs/tmp2
# 1.6G	/opt/homebrew/Caskroom/miniconda/base/envs/tmp2
```

tsibley: Thanks!

huddlej (approved):
I like this solution. We've started to use all of these tools for seasonal flu, now, too, so this is going to make everyone's life easier in the future.

My only note is that eventually this list will get so long that we'll be tempted to put it into a Conda environment YAML or even [its own Bioconda recipe](https://github.com/bioconda/bioconda-recipes/blob/25ee21573c577aa1ae899e0f7fbbc848d8a63865/recipes/nextstrain/meta.yaml) and we will have come full circle. Which I guess is a +1 for working on buildpacks (nextstrain/docker-base#71) sooner than later as a way to better manage our dependencies.

tsibley:
> My only note is that eventually this list will get so long that we'll be tempted to put it into a Conda environment YAML or even [its own Bioconda recipe](https://github.com/bioconda/bioconda-recipes/blob/25ee21573c577aa1ae899e0f7fbbc848d8a63865/recipes/nextstrain/meta.yaml) and we will have come full circle.

IIRC, the only reason we're not using the `nextstrain` meta-package in Bioconda for Conda-based installation (the original intent) is because the update path is unclear/difficult given limitations of `conda`/`mamba`?

One solution to that which I think we briefly considered but didn't pursue would be pinning versions of our packages in the meta-package and automating the update of the pins + making new meta-package releases. This is very much akin to the way to the Docker runtime image is managed currently, and there's lots of benefits to it…

(A reliable meta-package would also make it much easier to implement the vision I have for `nextstrain setup`, which would be a counterpart to `check-setup` that would do the work of getting a runtime (Docker, Conda, etc) setup. This is particularly compelling combined with the standalone installation archives for Nextstrain CLI to avoid bootstrapping issues.)

> Which I guess is a +1 for working on buildpacks ([nextstrain/docker-base#71](https://github.com/nextstrain/docker-base/issues/71)) sooner than later as a way to better manage our dependencies.

Buildpacks definitely help here for containerized runtimes (Docker, AWS Batch, Terra, (in-some-future) Singularity).

PR #119, opened 2022-07-18
Replace sphinx_copybutton with nextstrain.sphinx.theme

victorlin:

The new extension already has sphinx_copybutton pre-installed and configured.

## Related issue(s)

https://github.com/nextstrain/sphinx-theme/issues/30

## Testing

- [ ] copy button works on RTD preview build

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -6,7 +6,7 @@ dependencies:
   - make
   - pip
   - pip:
-    - nextstrain-sphinx-theme>=2020.2
+    - nextstrain-sphinx-theme>=2022.5
     - docutils<0.18
     - recommonmark
     - requests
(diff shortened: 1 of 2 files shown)


tsibley:
Should we update the dep declaration to `nextstrain-sphinx-theme >=2022.5` to be safe?

victorlin:
Oh good point!

victorlin:
Ok, should be good on all PRs now (and created a new one for CLI: https://github.com/nextstrain/cli/pull/199)

tsibley (approved):
Non-blocking, but I would squash these two commits together since they're part of the same change (one depends on the other and vice versa).

PR #120, opened 2022-07-19
install: Use --strict-channel-priority for Conda channels

tsibley:
This is a documented requirement by both conda-forge¹ and bioconda² so
that packages from the "defaults" channel get selected only if they're
not in either of the others.  It is also required to solve a dependency
resolution issue affecting Nextstrain CLI installed from Bioconda that
arises because of the otherwise "flexible" channel priority.³

While conda-forge and bioconda suggest setting this in .condarc with
`conda config`, I've used the command-line option instead to match our
convention of specifying channels on the command-line too.

Note that specifying `-c defaults` explicitly isn't necessary most of
the time because it should already be last priority in the channel list,
but it's possible folks have used .condarc to remove it entirely due to
outdated advice.

¹ https://conda-forge.org/docs/user/tipsandtricks.html#using-multiple-channels
² https://bioconda.github.io/#usage
³ https://github.com/nextstrain/cli/issues/184

### Testing
- [x] Verified new install commands work on Linux
- [x] Built docs locally to confirm includes work
- [x] Verify includes work on RTD preview build


corneliusroemer:
Thanks for this @tsibley 

Since people often update conda environments, it is important to write channels and strictness into the environment config file.

That way one does not need to add all the channel flags for every update.

Maybe we could add this as an optional step if you feel it would make the installation otherwise seem to complicated.

src/install.rst
```diff
@@ -73,9 +73,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
                1. Install Docker Desktop using `the official guide <https://docs.docker.com/desktop/mac/install/>`_.
                2. Create a conda environment named ``nextstrain`` and install the Nextstrain CLI:
 
-                  .. code-block:: bash
-
-                     mamba create -n nextstrain -c conda-forge -c bioconda nextstrain-cli --yes
+                  .. include:: snippets/conda-install-minimal.rst
```
corneliusroemer, line 77: Add something along the lines of

```suggestion
conda config --env --set channel_priority strict
conda config --env --add channels conda-forge
conda config --env --add channels bioconda
```

to have the environment set up once and for all and being able to update without having to add flags.

Otherwise things will break if people update and forget a flag. We could say "optional" here, but I would really like this to be mentioned in the docs. It's just 1 extra copy/paste and adds a lot of robustness.

corneliusroemer: In fact we even suggest this type of *future-proofing* for apple silicon, so I don't see this as breaking with convention, and convention shouldn't prevent us from encouraging robustness:
>  Ensure future Conda commands in this environment use Intel packages too.
> ```conda config --env --set subdir osx-64```

corneliusroemer: I would add this as a snippet so it can be reused in many places (as it should be, whenever one sets up an environment this should be run - if not included in ~/.condarc in general)

corneliusroemer (changes requested):
I'd really like for the `conda config` stuff to be included in docs

src/install.rst
```diff
@@ -73,9 +73,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
...
```
victorlin, line 76: _(non-blocking)_

Nice, learned a new reST feature!

src/snippets/conda-install-full.rst
```diff
@@ -0,0 +1,5 @@
+.. code-block:: bash
+
+   mamba create -n nextstrain \
+     -c conda-forge -c bioconda -c defaults --strict-channel-priority --yes \
+     nextstrain-cli augur auspice nextalign nextclade snakemake git epiweeks pangolin pangolearn
```
victorlin, line 5: I'm all in favor for less duplication, but this won't work in Anaconda Prompt (cmd or PS) where the line continuation char is `` ` `` instead of `\`.

Suggestion: rename this file to `src/snippets/conda-install-full-bash.rst` and have another file `src/snippets/conda-install-full-powershell.rst` with contents:

```rst
.. code-block:: powershell

   mamba create -n nextstrain `
     -c conda-forge -c bioconda -c defaults --strict-channel-priority --yes `
     nextstrain-cli augur auspice nextalign nextclade snakemake git epiweeks pangolin pangolearn
```

(same for `conda-install-minimal.rst`)

src/install.rst
victorlin: I'm also in favor of adding this since:

1. We have the code block copy button, so it's not much extra work on the user.
2. It's only an extra step for the native runtime setup, doesn't affect Docker runtime.
3. I'd expect long-term users of the native runtime to eventually want to use this config either in the env or `~/.condarc`.

victorlin (changes requested):

src/install.rst
tsibley: Sure, that's fine. I was trying to avoid a larger rework of these install commands/instructions, but I agree it's more robust to set in config and agree that it's likely desirable long-term for users.

That said, we use `config config --env` for the M1 users instead of modifying the global config. Should we do that for channels too so we don't modify their global config?

I repushed just now to add the command-line options to the `mamba update` invocation, which was missed, but this will go away with the use of config.

src/snippets/conda-install-full.rst
tsibley: Ah, well, that's just the code that was there. But I can fix it in this PR too.

tsibley: Ah, nevermind, I indeed missed where it went from `.. code-block:: none` to `.. code-block:: bash`!

victorlin: Oh, actually this suggestion would just be for `conda-install-minimal.rst` since the full install for Native isn't a valid installation path for Windows.

tsibley:
@victorlin @corneliusroemer So ultimately I think what we're proposing here is having users run these commands:

    mamba create -n nextstrain
    conda activate nextstrain

    conda config --env --add channels defaults
    conda config --env --add channels bioconda
    conda config --env --add channels conda-forge
    conda config --env --set channel_priority strict

    mamba install --yes nextstrain-cli …

What are your thoughts on having this as a single command block in a single installation step vs. split into three blocks in three installation steps?

victorlin:
I think a single command block with comments should be good, similar to this:

https://github.com/nextstrain/docs.nextstrain.org/blob/730c1d5a5f9fa4dea74464b1ce9352aa518d5876/src/install.rst?plain=1#L92-L101

src/install.rst
corneliusroemer: Sounds great! Thanks for thorough checking and thinking through!

corneliusroemer:
There's ups and downs but I think I prefer the single block too

Downside: someone might copy only the install line, missing the important channel settings (I can't see in the proposal whether we scrap channel specs in the install line or not).

We could add redundancy there at the risk of having duplications.

We could condense creation and installation into one line, then have one line to make channel settings permanent. Downside is duplication.

Up to you guys, I'd be in favour of very briefly explaining why channel settings matter as this may be new to people. I've never done strict channel etc through config till now

tsibley:
Repushed. I left the addition of the command-line options and then added another commit to convert them to env config because it required a bit of restructuring and I didn't want the addition of the options/config to be lost in the noise.  

> Downside: someone might copy only the install line, missing the important channel settings (I can't see in the proposal whether we scrap channel specs in the install line or not).
> 
> We could add redundancy there at the risk of having duplications.

Channels are scrapped from command-line args since they're in config and it vastly simplifies the commands. I'm not worried about guessing if someone only copies the first line or not. We could repeat them, but I don't think it's worth it (unless we see a spate of this kind of issue!).

src/snippets/conda-install-full.rst
tsibley: Done in ea1af71.

src/install.rst
tsibley: Done in dc1b9a5.

corneliusroemer (approved):
Really good! Let's merge this, it's so much better than anything we have right now.

Do we have a PR test build anywhere I could quickly check out?

One additional thing I could imagine being useful having added would be a short script to install miniconda in a scripted way. Every time I set up a computer I have to sift through blog posts to find all the commands I need.

If I as a not-unexperienced conda user struggle, I can only imagine the confusion if you're new to the whole eco system. Many people will have conda already, but it would really be nice if we could have a few lines to setup from scratch for new users.

tsibley:
Yes, you can see preview at: https://nextstrain--120.org.readthedocs.build/en/120/install.html

You can find this link in the GitHub checks section at the bottom of the PR:
image

As for a short script to download and run Miniconda, we provide a multi-command example of doing this for Linux (incl. WSL), but not for macOS or Windows:

image

Not sure why the difference, but possibly because we're suggesting the graphical installers for macOS and Windows. In any case, addressing that is a separate PR in my mind and not one I'm going to do right now.

corneliusroemer:
Thanks! For macOS it's very similar so would be good to also suggest there. I'll make an issue.

PR #122, opened 2022-07-25
Install nits

tsibley:
Various nits I noticed today in the installation instructions. See commit messages for details. I think these are likely to be uncontroversial, but I'll wait for review to make sure.

### Testing
- [x] CI passes
- [x] Renders well locally

### DIFF
--- a/src/install.rst
+++ b/src/install.rst
@@ -70,7 +70,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
                   If using a newer Mac with an `Apple silicon chip <https://support.apple.com/en-us/HT211814>`_ (e.g. M1), **Native** installation is recommended due to slowness with the Docker installation. `We are considering ways to improve this <https://github.com/nextstrain/docker-base/issues/35>`_.
 
-               1. `Install Docker Desktop using the official guide <https://docs.docker.com/desktop/mac/install/>`_.
+               1. `Install Docker Desktop using the official guide <https://docs.docker.com/desktop/install/mac-install/>`_.
                2. Create a Conda environment named ``nextstrain``:
 
                   .. include:: snippets/conda-create-bash.rst
--- a/src/install.rst
+++ b/src/install.rst
@@ -275,7 +275,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
                   .. code-block:: bash
 
-                     sudo usermod -aG docker $USER
+                     sudo gpasswd --add $USER docker
 
                3. Restart your machine.
                4. Create a Conda environment named ``nextstrain``:
--- a/src/install.rst
+++ b/src/install.rst
@@ -262,14 +262,15 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
             .. group-tab:: Docker (recommended)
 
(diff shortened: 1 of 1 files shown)


src/install.rst
```diff
@@ -70,7 +70,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
                   If using a newer Mac with an `Apple silicon chip <https://support.apple.com/en-us/HT211814>`_ (e.g. M1), **Native** installation is recommended due to slowness with the Docker installation. `We are considering ways to improve this <https://github.com/nextstrain/docker-base/issues/35>`_.
 
-               1. Install Docker Desktop using `the official guide <https://docs.docker.com/desktop/mac/install/>`_.
+               1. `Install Docker Desktop using the official guide <https://docs.docker.com/desktop/mac/install/>`_.
```
j23414, line 73: (non-blocking comment)

When I tried to get to: https://docs.docker.com/desktop/mac/install/
I got redirected to: https://docs.docker.com/desktop/install/mac-install/

Which is fine, or feel free to link it directly.

tsibley: Switched in f29c12d.

src/install.rst
```diff
@@ -270,11 +273,19 @@ These instructions will install the Nextstrain CLI and tools to run and view you
                      # follow through installation prompts
                      rm get-docker.sh
 
+                  .. note::
+
+                     Alternatively, you can install Docker Engine using your system's package manager:
+
+                     .. code-block:: bash
+
+                        sudo apt install docker.io
```
victorlin, line 282: Cool, wasn't aware of this option since it wasn't advertised on https://docs.docker.com/engine/install/ubuntu/. Any reason not to have this as the only option to keep things simple? With a more general note like:

> See [Docker's installation documentation](https://docs.docker.com/engine/install/ubuntu/) for alternative installation methods.

victorlin: Oh wait, that page actually says:

> Older versions of Docker were called `docker`, `docker.io`, or `docker-engine`. If these are installed, uninstall them

So maybe we shouldn't advertise this method?

victorlin (changes requested):
Thanks for cleaning this up! Just one small thing.

src/install.rst
tsibley: `docker.io` is the package provided by the Ubuntu/Debian maintainers.

`docker-ce` and co. are the packages provided by the Docker maintainers.

The latter will generally be the latest versions of Docker while the former will lag per Ubuntu/Debian's typical stability policies. OTOH, the latter requires setup of Docker's third-party package repos, signing keys, etc. while the former does not. Since we don't need the latest and greatest version, it seems simpler to use the Ubuntu-provided packages. Docker instructs you to uninstall `docker.io`, etc. because you can't have both Ubuntu's package and Docker's package installed, and `apt` will report a conflict if you try to.

tsibley: All that said, I'd be in favor of making the simplest option the most prominent as you initially suggested.

tsibley: Repushed, replacing a548356 with 21aa2a4.

victorlin (approved):

PR #123, opened 2022-07-26
Use standalone installation of Nextstrain CLI for Docker runtimes

tsibley:
[Preview](https://nextstrain--123.org.readthedocs.build/en/123/install.html)

---

Reorganizes the tabs to be runtime first then OS (instead of OS then runtime) as I think this makes more sense now that the initial steps of installing and setting up a base Conda env aren't shared.  It also means we don't have multiple levels of ordered lists (steps), which is a nice simplification.
The goal is simplify this installation path quite a bit by avoiding the Python bootstrapping issue.

Reorganizes the tabs to be runtime first then OS (instead of OS then runtime) as I think this makes more sense now that the initial steps of installing and setting up a base Conda env aren't shared.  It also means we don't have multiple levels of ordered lists (steps), which is a nice simplification.

Moves verbose explanatory information out into the FAQ and adds more explanation about how to make various choices.

Moves the hint for WSL on Windows installs next to the hint for Native installs at the bottom.  This avoids duplicating it in both of the WSL on Windows tabs and it taking up a lot of visual space in those tabs.

Related-to: <https://github.com/nextstrain/nextstrain.org/pull/582>
Related-to: <https://github.com/nextstrain/cli/pull/217>

## Todo

- [x] Actual install commands and nextstrain.org endpoints serving the installer
- [x] ~Produce MSIs for Windows~ Nope, PowerShell program works just fine. Maybe we'll do MSIs later.
- [x] Document how to make various choices (runtime, WSL or not)
- [x] Figure out best approach for `PATH` handling on Unix (e.g. auto-add to shell config? suggest sourcing a provided file for your shell?)
- [x] Release new Nextstrain CLI with `init-shell` for new standalone installers to use

## Related

- https://github.com/nextstrain/cli/pull/217
- https://github.com/nextstrain/nextstrain.org/pull/582

### DIFF
--- a/src/install.rst
+++ b/src/install.rst
@@ -271,21 +271,37 @@ Update an existing installation
 Troubleshoot a broken installation
 ==================================
 
-If Conda fails to install or update Nextstrain using the commands above, it's possible that Conda itself is out-of-date or that Conda cannot figure out how to resolve the environment's dependencies.
-Try the following approaches, to fix these broken installations.
+.. tabs::
 
-Remove your environment and start from scratch
-----------------------------------------------
+   .. group-tab:: Docker
 
-Starting from scratch often fixes problems with Conda environments.
-To start over with a new Nextstrain environment, delete your current environment.
+      The Docker runtime requires that the Docker service is running on your computer behind the scenes.
+      If you see a message like::
 
-.. code-block:: bash
+         Cannot connect to the Docker daemon at […]. Is the docker daemon running?
 
-    conda activate base
-    conda env remove -n nextstrain
+      Then it is likely that the Docker service is not running.
+      On macOS and Windows, try quitting Docker Desktop (if it's open) and restarting it.
+      On Linux, try running ``sudo systemctl restart docker``.
 
-Then, repeat the installation instructions above, starting with the update of Conda itself.
+      Running ``nextstrain check-setup`` will also report potential issues.
+      Make sure there are no errors or warnings reported for the Docker runtime.
+
+
+   .. group-tab:: Native
+
+      If Conda fails to install or update Nextstrain using the commands above, it's possible that Conda itself is out-of-date or that Conda cannot figure out how to resolve the environment's dependencies.
+      Starting from scratch often fixes problems with Conda environments.
+      To start over with a new Nextstrain environment, delete your current environment.
+
+      .. code-block:: bash
+
+          conda activate base
+          conda env remove -n nextstrain
+
+      Then, repeat the installation instructions above, starting with the update of Conda itself.
+
+If you the above isn't sufficient and you need more help troubleshooting, please post to our `discussion forum <https://discussion.nextstrain.org/c/help-and-getting-started/6>`__ where members of the community and the Nextstrain team can help out.
 
 Alternate installation methods
 ==============================
--- a/src/install.rst
+++ b/src/install.rst
@@ -6,342 +6,267 @@ Installing Nextstrain
 
     Before installing, we recommend you read about the :doc:`parts of Nextstrain </learn/parts>`.
 
-The following instructions describe how to install Nextstrain's software tools, including:
+The following instructions describe how to install the full suite of Nextstrain's software tools, including:
 
-  * Augur: a bioinformatics toolkit for the analysis of pathogen genomes
-  * Auspice: a tool for interactive visualization of pathogen evolution
-  * Nextstrain CLI: tools for management of analysis workflows and environments
+  * Augur, for bioinformatic analysis of pathogen genomes
+  * Auspice, for interactive visualization of pathogen evolution
+  * Nextstrain CLI, for management of analysis workflows and environments
 
-.. note::
+When completed, you'll be ready to run Nextstrain :term:`workflows <workflow>`.
 
-    If you want to :doc:`contribute to the development of Nextstrain </guides/contribute/index>` or if you prefer to manage your own custom environment (e.g., a Conda environment, Docker image, environment modules on a cluster, etc.), see the individual installation documentation for :doc:`Nextstrain CLI <cli:installation>`, :doc:`Augur <augur:installation/installation>`, and :doc:`Auspice <auspice:introduction/install>`.
-
-.. contents:: Table of Contents
-   :local:
-   :depth: 1
-
-
-Background
-==========
-
-`Conda <https://docs.conda.io/en/latest/>`_ is a package and environment management system that allows you to install Python and other software into controlled environments without disrupting other software you have installed (e.g., on your computer, your shared cluster, etc.).
-Conda provides an appropriate version of Python required by all approaches to installing Nextstrain tools. Miniconda is the minimal installation of the command-line interface to Conda.
-
-`Mamba <https://github.com/mamba-org/mamba>`_ is a drop-in replacement for most ``conda`` functionality that implements a faster dependency solving algorithm in C++ and multithreaded downloads.
-As a result, Mamba can install Conda packages much faster and more accurately than the original Conda installer.
-
-`Docker <https://docker.com/>`_ is a container system freely-available for all platforms.
-When you use Docker to run Nextstrain components, you don’t need to manage any other Nextstrain software dependencies as validated versions are already bundled into `a container image by the Nextstrain team <https://github.com/nextstrain/docker-base/>`_.
-
-Installation Steps
+Installation steps
 ==================
 
-These instructions will install the Nextstrain CLI and tools to run and view your own Nextstrain analyses. Configuration options vary by operating system.
+Steps vary by runtime option (Docker, native) and operating system (macOS, Windows, WSL on Windows, Linux).
+For help choosing, refer to our :doc:`/reference/faq`, such as:
+
+  * :ref:`what-are-docker-conda-mamba-wsl-etc`
+  * :ref:`choosing-a-runtime`
+  * :ref:`when-to-use-wsl`
+
+First, install a Nextstrain runtime.
 
 .. tabs::
 
-   .. group-tab:: macOS
+   .. group-tab:: Docker
 
-      .. note::
+      .. tabs::
 
-         If you are an experienced user, you can replace ``conda`` with ``pip`` but :doc:`note the extra installation steps for Augur <augur:installation/installation>` and :doc:`install Auspice via npm <auspice:introduction/install>`.
+         .. group-tab:: macOS
 
-      1. Install Miniconda:
+            .. warning::
 
-         .. The installer link is taken from https://docs.conda.io/en/latest/miniconda.html.
+               If using a newer Mac with an `Apple silicon chip <https://support.apple.com/en-us/HT211814>`_ (e.g. M1), the **native** runtime is recommended due to slowness with the Docker runtime. `We are considering ways to improve this <https://github.com/nextstrain/docker-base/issues/35>`_.
 
-         a. `Download the installer <https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.pkg>`_.
+            1. `Install Docker Desktop using the official guide <https://docs.docker.com/desktop/install/mac-install/>`_.
+            2. Install the Nextstrain CLI:
 
-            .. note::
+               .. code-block:: bash
 
-                  This is the Intel x86 64-bit installer, :ref:`which we recommend even for Mac computers with Apple silicon (e.g. M1) <why-intel-miniconda-installer-on-apple-silicon>`.
+                  curl -fsSL --proto '=https' https://nextstrain.org/cli/installer/mac | bash
 
-         b. Open the downloaded file and follow through installation prompts.
 
-      2. Open a terminal window.
-      3. Install Mamba on the ``base`` Conda environment:
+         .. group-tab:: Windows
 
-         .. code-block:: bash
+            1. `Install Windows Subsystem for Linux (WSL) 2 <https://docs.microsoft.com/en-us/windows/wsl/install>`_.
 
-            conda install -n base -c conda-forge mamba --yes
-            conda activate base
+               .. note:: You may have to restart your machine when configuring WSL.
 
-      4. Install the Nextstrain components. There are two options:
+            2. `Install Docker Desktop with WSL 2 backend <https://docs.docker.com/desktop/windows/wsl/>`_.
+            3. Open a PowerShell prompt (not a WSL prompt) as your own user (not as an administrator).
+            4. Install the Nextstrain CLI.
 
-         .. tabs::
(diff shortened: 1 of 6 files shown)


src/install.rst
```diff
@@ -24,328 +24,266 @@ The following instructions describe how to install Nextstrain's software tools,
...
+.. tabs::
 
-                  .. include:: snippets/conda-create-bash.rst
+   .. group-tab:: Docker (recommended)
 
-               2. Install all the necessary software:
+      .. tabs::
 
-                  .. include:: snippets/conda-install-full-bash.rst
+         .. group-tab:: macOS
```
victorlin, line 49: Nice, I'm on board with this. Seems appropriate in this PR to add some text describing how to choose between Docker vs. Native (would supersede #106) since this decision on the user is very obvious from the top-level tabs.

tsibley: Indeed!

src/install.rst
```diff
@@ -24,316 +24,266 @@ The following instructions describe how to install Nextstrain's software tools,
...
-            bash Miniconda3-latest-Linux-x86_64.sh
-            # follow through installation prompts
-            rm Miniconda3-latest-Linux-x86_64.sh
+         .. group-tab:: macOS
 
-      4. Install Mamba on the ``base`` Conda environment:
+            .. warning::
 
-         .. code-block:: bash
+               If using a newer Mac with an `Apple silicon chip <https://support.apple.com/en-us/HT211814>`_ (e.g. M1), first run these commands to ensure Conda creates the environment with ``osx-64`` emulation:
```
tsibley, line 139: This no longer makes sense in its new location given that installation of Miniconda (`conda`) doesn't happen now until *after* this warning. Need to re-arrange, but also, @victorlin has a better alternative suggestion of using the x86_64 installer for Miniconda always, even on M1 hardware.

tsibley:
With some more tinkering with Windows and PowerShell, I think we can support the following as the actual commands the user runs for Linux, macOS, and Windows:

```sh
curl -fsSL --proto '=https' https://nextstrain.org/cli/installer/linux | bash
```

```sh
curl -fsSL --proto '=https' https://nextstrain.org/cli/installer/mac | bash
```

```pwsh
Invoke-RestMethod https://nextstrain.org/cli/installer/windows | Invoke-Expression
```

The first two URLs will resolve to the same Bash script and the second will resolve to a PowerShell script. Both scripts will be hosted in the Nextstrain CLI git repo and resolved by nextstrain.org using redirects.

Things I noted about PowerShell:

- We could use actual `curl` if **PowerShell Core 7+** is installed, but Windows 10 (what I'm testing with) comes with **Windows PowerShell 5**. I think better to use `Invoke-RestMethod` and avoid requiring a PowerShell upgrade, even if it means we don't get `curl`'s better connection security guarantees.
- While looking into PowerShell upgrades just now, I noted https://aka.ms/install-powershell.ps1 which is a) much more comprehensive than we need but b) also full of nice things to potentially cherry pick into the PowerShell script I wrote and have been honing.

tsibley:
Example of installing on Windows:

win-install

Works great with current script!

src/install.rst
tsibley: [Relevant notes](https://docs.google.com/document/d/1dODYr-3T8LAQt-ewMZKNrcy1Wvy70Fw-h8exXbIwbyM/edit#heading=h.7nuap5p45ey7) (also linked from the linked PR).

src/install.rst
tsibley: Resolved by rebasing this PR onto the latest master and incorporating the changes there addressing this since I initially branched.

src/install.rst
tsibley: Added in 9f61ba8.

src/install.rst
```diff
@@ -6,342 +6,267 @@ Installing Nextstrain
...
 
-            Setting default environment to <option>.
+            1. `Install Docker Desktop using the official guide <https://docs.docker.com/desktop/install/mac-install/>`_.
+            2. Install the Nextstrain CLI:
 
-   .. group-tab:: Windows
+               .. code-block:: bash
 
-      .. note::
+                  curl -fsSL --proto '=https' https://nextstrain.org/cli/installer/mac | bash
```
victorlin, line 46: With an installation script that does things automatically, it would be good to include instructions on how to uninstall. However, it can get just as complex as the installation docs itself with the various installation paths (current and past). Maybe something along these lines in the FAQ page would be a good start:

1. If `~/.local/share/nextstrain`/`~\AppData\Local\nextstrain` exists, remove it.
2. If everything is in a Conda environment, remove the Conda environment.
3. If Docker was used, remove `nextstrain/*` Docker images.
4. Optionally, uninstall Docker/Conda/WSL if only used for Nextstrain.

I don't think we should worry about installations on an independently managed environment (e.g. `pip install nextstrain-cli`).

In the longer term, maybe a `nextstrain uninstall` command would be better.

src/install.rst
```diff
@@ -6,342 +6,267 @@ Installing Nextstrain
...
+For help choosing, refer to our :doc:`/reference/faq`, such as:
 
-               1. `Install Docker Desktop using the official guide <https://docs.docker.com/desktop/install/mac-install/>`_.
-               2. Create a Conda environment named ``nextstrain``:
+  * :ref:`what-are-docker-conda-mamba-wsl-etc`
+  * :ref:`choosing-a-runtime`
+  * :ref:`when-to-use-wsl`
 
-                  .. include:: snippets/conda-create-bash.rst
+First, install the Nextstrain components.
```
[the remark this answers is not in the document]
tsibley: Good point. Addressed and [repushed](https://github.com/nextstrain/docs.nextstrain.org/compare/52b7abda2cbe90139037060f9aa5259aa9744d23..fa63abb679b4830cc82770666b9d8a887fc3cf7f).

src/install.rst
```diff
@@ -6,342 +6,267 @@ Installing Nextstrain
...
-
-                  .. include:: snippets/conda-create-bash.rst
-
-               2. Install all the necessary software:
-
-                  .. include:: snippets/conda-install-full-bash.rst
+.. tabs::
 
-      5. Confirm that the installation worked.
+   .. group-tab:: Docker (recommended)
```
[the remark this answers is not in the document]
tsibley: :+1: Addressed and [repushed](https://github.com/nextstrain/docs.nextstrain.org/compare/52b7abda2cbe90139037060f9aa5259aa9744d23..fa63abb679b4830cc82770666b9d8a887fc3cf7f).

src/install.rst
tsibley: Ah, good point. I agree. Although instead of an FAQ entry, I'd think to make an "Uninstalling" section at the bottom of the install page, where it'll be more visible and not require a link out from install to FAQ.

tsibley: Pushed as [77424e3](https://github.com/nextstrain/docs.nextstrain.org/pull/123/commits/77424e399b64a016c140009b7594da99b29ff11e).

tsibley:
Pushed two other minor improvements as well with the new Uninstall section.

src/install.rst
```diff
@@ -6,361 +6,320 @@ Installing Nextstrain
...
+               If using a newer Mac with an `Apple silicon chip <https://support.apple.com/en-us/HT211814>`_ (e.g. M1), the **native** runtime is recommended due to slowness with the Docker runtime. `We are considering ways to improve this <https://github.com/nextstrain/docker-base/issues/35>`_.
 
-               3. Install the Nextstrain CLI:
+            1. `Install Docker Desktop using the official guide <https://docs.docker.com/desktop/install/mac-install/>`_.
+            2. Install the Nextstrain CLI:
 
-                  .. code-block:: bash
+               .. code-block:: bash
 
-                     mamba install --yes nextstrain-cli
```
victorlin, line 88: Since the setup of a `nextstrain` Conda environment is no longer common across all runtimes, it'd make sense to also reword subsequent tutorials:

- 6b032ccb20cc1ad7551cfd3fe60438c6de8f0fa2 (add to this PR)
- https://github.com/nextstrain/ncov/commit/2016068a3436aaca5aec55b4c339c066fa5a04c8 (create external PR)

src/install.rst
```diff
@@ -6,361 +6,320 @@ Installing Nextstrain
...
-            nextstrain check-setup --set-default
+         mamba update -n base conda mamba
+         conda activate nextstrain
+         mamba update --all
 
-         The final output from the last command should look like this, where ``<option>`` is the option chosen in the previous step:
 
-         .. code-block:: none
+Troubleshoot a broken installation
+==================================
```
victorlin, line 272: This whole section is Conda-specific, so maybe preface it for "native" runtime only. Or, add tabs to be consistent with other sections on this page, and add troubleshooting steps for standalone CLI / Docker runtime (if any). Some that come to mind:

- Run `nextstrain check-setup`
    - Check that `nextstrain-cli is up to date`
    - Check that there are no warnings for the Docker runtime

src/reference/faq.rst
```diff
@@ -9,6 +9,62 @@ Installation
...
+The two runtimes provide the same experience running Nextstrain workflows with ``nextstrain build`` (and most other ``nextstrain`` commands) but vary in installation and update methods and predictability/stability over time.
+
+There's not one right answer for everyone or every situation.
+That's why we provide both options (and will potentially provide more in the future).
+The Nextstrain team uses both runtimes extensively.
+
+Your preference is perhaps the best reason to choose one vs. the other.
+
+   - If you have a preference for containers, then choose our Docker runtime.
+   - If you have a preference for Conda, then choose our native runtime.
```
victorlin, line 44: Refer to runtimes consistently:

```suggestion
   - If you have a preference for containers, then choose the Docker runtime.
   - If you have a preference for Conda, then choose the native runtime.
```

victorlin (changes requested):

src/install.rst
tsibley: Ah, yes! I added the first to this PR. Will you open an external PR for the second since it's your change?

src/install.rst
tsibley: Good point. I added tabs in 0a560ef.

tsibley:
Added one additional commit, [45fb0f3](https://github.com/nextstrain/docs.nextstrain.org/pull/123/commits/45fb0f33725c82eb9d3e709fc004629369b91283), while revisiting the page to do the troubleshooting work above.

src/reference/faq.rst
tsibley: Committed as 21b65c3. I can fix it up into the first commit in this PR if you'd like.

tsibley:
and added ec7d132 to fix the CI failures unrelated to this PR. If this PR doesn't merge today, we should cherry-pick ec7d132 to master sooner.

src/install.rst
victorlin: Opened: https://github.com/nextstrain/ncov/pull/1002

src/reference/faq.rst
victorlin: Sure, that sounds reasonable.

src/install.rst
```diff
@@ -6,361 +6,329 @@ Installing Nextstrain
...
-      5. Install the Nextstrain components. There are two options:
 
-         .. tabs::
+         .. group-tab:: Windows
 
-            .. group-tab:: Docker (recommended)
+            .. note::
 
-               1. On Windows, `install Docker Desktop with WSL 2 backend <https://docs.docker.com/desktop/windows/wsl/>`_.
+               Due to installation constraints, there is no way to use the native Nextstrain runtime on Windows directly. Follow steps for **WSL on Windows** if the native environment is desired, or use the **Docker**-based steps instead.
```
victorlin, line 151: Consistently reference native runtime:

```suggestion
               Due to installation constraints, there is no way to use the native runtime on Windows directly. Follow steps for **WSL on Windows** if the native runtime is desired, or use the **Docker**-based steps instead.
```

victorlin (approved):
Great stuff! This seems much easier to follow now.

- [x] Standalone installer testing https://github.com/nextstrain/cli/pull/217#pullrequestreview-1109924895
- [x] Read through the other changed sections

src/install.rst
```diff
@@ -6,361 +6,329 @@ Installing Nextstrain
...
 
-                  .. include:: snippets/conda-create-bash.rst
+Troubleshoot a broken installation
+==================================
 
-               2. Install all the necessary software:
+.. tabs::
 
-                  .. include:: snippets/conda-install-full-bash.rst
+   .. group-tab:: Docker
```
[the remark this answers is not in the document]
tsibley: Yep! It's much more noticeable now that the page is slimmer and there are more tabbed sections.

src/install.rst
tsibley: Fixed up in 35698d0.

src/install.rst
```diff
@@ -6,361 +6,329 @@ Installing Nextstrain
...
 
-      7. Install the Nextstrain CLI:
+            3. Restart your machine.
+            4. Install the Nextstrain CLI:
 
-         .. code-block:: powershell
+               .. code-block:: bash
 
-            mamba install --yes nextstrain-cli
+                  curl -fsSL --proto '=https' https://nextstrain.org/install/cli/linux | bash
```
victorlin, line 109: This isn't the right URL, but the others are correct.

```suggestion
                  curl -fsSL --proto '=https' https://nextstrain.org/cli/installer/linux | bash
```

victorlin (changes requested):

src/install.rst
tsibley: Ah, good catch! Fixed.

tsibley:
[Repushed](https://tsibley.net/paste/2022-10-06Kh72v9_Q) to rebase onto latest `master` and:

1. Adjust the uninstall steps now that the standalone archive is installed into `~/.nextstrain`,
2. Drop 19f0553 since an equivalent change was already applied to `master`,
3. Call out the final shell config setup step so users don't miss it (7e51fcd).

With that, this PR is ready to go but must wait for a non-development release of Nextstrain CLI so the `init-shell` command (used by the installers) is available.

tsibley:
With the 5.0.0.dev0 pre-release just a moment ago, you can test the installation commands in this PR on macOS and Linux by appending `-s 5.0.0.dev0` to the end, e.g.:

```
curl -fsSL --proto '=https' https://nextstrain.org/cli/installer/linux | bash -s 5.0.0.dev0
```

On Windows, you'll have to slightly adjust the command to:

```
Invoke-Expression "& { $(Invoke-RestMethod https://nextstrain.org/cli/installer/windows) } 5.0.0.dev0"
```

tsibley:
Releasing Nextstrain CLI 5.0.0 shortly and then will merge this.

Issue #121, opened 2022-07-25
DOC: Add miniconda install scripts to macOS (and maybe Windows) docs


corneliusroemer:
Let's add short scripts for installing Miniconda on macOS (and if possible Windows) as we do for Linux

>As for a short script to download and run Miniconda, we provide a multi-command example of doing this for Linux (incl. WSL), but not for macOS or Windows:

>![image](https://user-images.githubusercontent.com/79913/180834865-c5b747e0-9fef-4b90-bca1-e7edd62dbe5d.png)

>Not sure why the difference, but possibly because we're suggesting the graphical installers for macOS and Windows. In any case, addressing that is a separate PR in my mind and not one I'm going to do right now.

_Originally posted by @tsibley in https://github.com/nextstrain/docs.nextstrain.org/issues/120#issuecomment-1194364569_

j23414:
There are several installer scripts for Mac (Intel 86 vs Apple M1 chip) listed and linked in this [miniconda table](https://docs.conda.io/en/latest/miniconda.html#latest-miniconda-installer-links). 

For **Mac Intel 86** it can look like:

```
wget https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.sh
bash Miniconda3-latest-MacOSX-x86_64.sh
# follow through installation prompts
rm Miniconda3-latest-MacOSX-x86_64.sh
```

While for **Mac Apple M1 chip** it can look like:

```
wget https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-arm64.sh
bash Miniconda3-latest-MacOSX-arm64.sh
# follow through installation prompts
rm Miniconda3-latest-MacOSX-arm64.sh
```

@corneliusroemer would documenting both on the MacOS be acceptable? Or do you have a sense that most people have moved to the M1 chip on Mac?

The Window's installer scripts mainly pull exe files

```
wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Windows-x86_64.exe
wget https://repo.anaconda.com/miniconda/Miniconda3-latest-Windows-x86.exe
```

j23414:
We could also make it more general but complicated to read ;)

```
# Pick one OS and ARCH pair
OS=Windows
ARCH=x86_64.exe
ARCH=x86.exe

OS=MacOSX
ARCH=x86_64.sh
ARCH=arm64.sh

[...]

PR #124, opened 2022-08-05
Use Intel Miniconda installer for Apple silicon, start an installation FAQ page

victorlin:
_[(preview)](https://nextstrain--124.org.readthedocs.build/en/124/install.html#installation-steps)_

See FAQ contents for the reasoning behind using the Intel Miniconda
installer on Apple silicon.

### Related issue(s)

- Ideas from https://github.com/nextstrain/docs.nextstrain.org/issues/121#issuecomment-1197224332 and further discussion during a meeting with @tsibley @huddlej @j23414.

### Testing

- [x] checks pass

### DIFF
--- a/src/install.rst
+++ b/src/install.rst
@@ -48,9 +48,15 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
       1. Install Miniconda:
 
-         a. Go to the `installation page <https://docs.conda.io/en/latest/miniconda.html>`_.
-         b. Scroll down to the **Latest Miniconda Installer Links** section and click the MacOSX platform link that ends with **pkg**.
-         c. Open the downloaded file and follow through installation prompts.
+         .. The installer link is taken from https://docs.conda.io/en/latest/miniconda.html.
+
+         a. `Download the installer <https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.pkg>`_.
+
+            .. note::
+
+                  This is the Intel x86 64-bit installer, :ref:`which we recommend even for Mac computers with Apple silicon (e.g. M1) <why-intel-miniconda-installer-on-apple-silicon>`.
+
+         b. Open the downloaded file and follow through installation prompts.
 
       2. Open a terminal window.
       3. Install Mamba on the ``base`` Conda environment:
@@ -82,21 +88,6 @@ These instructions will install the Nextstrain CLI and tools to run and view you
                      mamba install --yes nextstrain-cli
 
             .. group-tab:: Native
-
-               .. warning::
-
-                  If using a newer Mac with an `Apple silicon chip <https://support.apple.com/en-us/HT211814>`_ (e.g. M1), first run these commands to ensure Conda creates the environment with ``osx-64`` emulation:
-
-                  .. code-block:: bash
-
-                     # Create a new environment using Intel packages called base_osx-64
-                     CONDA_SUBDIR=osx-64 conda create -n base_osx-64 python
-
-                     # Activate new Intel-based environment
-                     conda activate base_osx-64
-
-                     # Ensure future Conda commands in this environment use Intel packages too.
-                     conda config --env --set subdir osx-64
 
                1. Create a Conda environment named ``nextstrain``:
 
--- a/src/reference/faq.rst
+++ b/src/reference/faq.rst
@@ -1,0 +1,29 @@
+================================
+Frequently Asked Questions (FAQ)
+================================
+
+.. TODO: Add a table of contents once there are more sections. Doesn't seem necessary at the time of writing.
+
+Installation
+============
+
+There are many ways to install Nextstrain, and we aim to simplify the installation guide so it is easy to follow along. Here, you will find answers to some common questions about the installation process.
+
+.. _why-intel-miniconda-installer-on-apple-silicon:
+
+Why recommend the Intel Miniconda installer for Mac computers with Apple silicon (e.g. M1)?
+-------------------------------------------------------------------------------------------
+
+.. note::
+
+   `How to tell if your Mac has an Apple silicon chip <https://support.apple.com/en-us/HT211814>`_
+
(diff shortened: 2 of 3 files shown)


victorlin:
Tests are failing since the FAQ page isn't in any `toctree`, but [the preview is available](https://nextstrain--124.org.readthedocs.build/en/124/install.html#installation-steps) and ready for review.

corneliusroemer:
Thanks @victorlin 

Silly question, maybe I'm digging to deep, but I still don't understand what happens if you don't use the osx_64 setting: will I not get any mafft?

Can I somehow use osx_64 Emulation only for incompatible packages. So if arm Python exists, can I use it while installing emulated mafft? Or is this exactly what's happening?

I'd expect (naively) for conda to be able to use different archs like channels, with priorities.

But maybe I miss something fundamental. I should just try installing without the emulation thing ;)

victorlin:
@corneliusroemer just try running `mamba create -n tmp mafft` from your base environment (assuming you haven't set `subdir=osx_64` there). Output should look something like this:

```
Looking for: ['mafft']

pkgs/main/osx-arm64      [====================] (00m:00s) Done
bioconda/osx-arm64       [====================] (00m:00s) No change
pkgs/r/noarch            [====================] (00m:00s) No change
pkgs/r/osx-arm64         [====================] (00m:00s) No change
pkgs/main/noarch         [====================] (00m:00s) Done
conda-forge/osx-arm64    [====================] (00m:01s) Done
bioconda/noarch          [====================] (00m:01s) Done
conda-forge/noarch       [====================] (00m:02s) Done
Encountered problems while solving:
  - nothing provides requested mafft
```

As far as I'm aware, you have to explicitly set `subdir=osx_64` for Conda to look for packages under `<channel>/../osx-64`. I'm not sure how a mix of `osx-64` and `osx-arm64` installed packages would behave, but I'd expect some weirdness there or at least difficulty in managing the Conda environment.

src/install.rst
```diff
@@ -48,9 +48,15 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
       1. Install Miniconda:
 
-         a. Go to the `installation page <https://docs.conda.io/en/latest/miniconda.html>`_.
-         b. Scroll down to the **Latest Miniconda Installer Links** section and click the MacOSX platform link that ends with **pkg**.
-         c. Open the downloaded file and follow through installation prompts.
+         .. The installer link is taken from https://docs.conda.io/en/latest/miniconda.html.
+
+         a. Download the ``.pkg`` installer by clicking `here <https://repo.anaconda.com/miniconda/Miniconda3-latest-MacOSX-x86_64.pkg>`_.
```
tsibley: Pushed b9aa8bafed63dcafb17d61bebd070e4deffeb326 to remove the "click here" wording.

Dropped mention of `.pkg` entirely since we're directly linking the thing we want folks to install.

tsibley (changes requested):
Looks great!

I made a small wording tweak to avoid "click here" (pushed directly as a `fixup!` commit). If you agree, please rebase the fixup away into your commit.

As for where to link the installation FAQ, I'd suggest one of the following:

1. an "Installation FAQ" entry under "How-to Guides", at the bottom after "Contributing"
2. an "Installation FAQ" entry under "Reference" (which is not purely "reference" in the strict sense of Divio's Documentation System), maybe between "Data formats" and "Augur" or maybe just before "Style guide".
3. an "Installation" entry under a new (mostly prospective) "FAQs" section between the "How-to Guides" and "Reference" sections

I also suppose that an installation-specific FAQ may be a bit premature yet and it'd be better to have a single general "FAQ" page with an "Installation" heading (and more headings to come later). This page would slot into spots 1 or 2 as a single entry or in 3 with page headings included in the sidebar for that "FAQ" toctree.

victorlin:
@tsibley

Good point about the premature installation-specific FAQ page. I opted for a single FAQ page with an installation section, under **Reference**. I don't see FAQ pages fitting anywhere under Divio's Documentation System. Choosing without creating a new section, I think **Reference** seems more appropriate than **How-to Guides** since this FAQ page won't have steps to follow, and it's at least _referenced_ by other parts of our documentation.

src/install.rst
victorlin: Thanks, makes sense to me!

src/reference/faq.rst
```diff
@@ -0,0 +1,29 @@
...
+.. TODO: Add a table of contents once there are more sections. Doesn't seem necessary at the time of writing.
+
+Installation
+============
+
+There are many ways to install Nextstrain, and we aim to simplify the installation guide so it is easy to follow along. Here, you will find answers to some common questions about the installation process.
+
+.. _why-intel-miniconda-installer-on-apple-silicon:
+
+Why recommend the Intel Miniconda installer for Mac computers with Apple silicon (e.g. M1)?
```
victorlin, line 14: Reworded to reference "Apple silicon" instead of M1 for future-proofing.

tsibley: Yes, good point!

src/reference/faq.rst
```diff
@@ -0,0 +1,29 @@
...
+-------------------------------------------------------------------------------------------
+
+.. note::
+
+   `How to tell if your Mac has an Apple silicon chip <https://support.apple.com/en-us/HT211814>`_
+
+Apple silicon chips are great and efficient. However, many existing packages have not yet added support to run on this chip natively. An easy way to identify support on the `Bioconda packages page <https://anaconda.org/bioconda>`_ is to look for ``noarch`` or ``osx-arm64`` under the **Installers** section of a package. Without any of those, a package is not able to be installed natively on Apple silicon. This is the case for packages such as `MAFFT <https://anaconda.org/bioconda/mafft>`_ (a dependency of :term:`Augur`) and many other bioinformatics packages. For this reason, using an Apple silicon Miniconda installation for the average bioinformatics researcher can result in a difficult experience. To circumvent this, one may enable ``osx-64`` emulation using the ``subdir`` config on each Conda environment, but it is easier to install with emulation by default.
```
tsibley, line 21: s/this chip/these chips/ now that it's "Apple silicon chips" instead of "The M1 chip".

tsibley (approved):

src/reference/faq.rst
victorlin: Good catch. Done: 0572e6200f3c2208a488c3c683eb14b222ce9cdd

src/install.rst
```diff
@@ -83,21 +89,6 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
             .. group-tab:: Native
 
-               .. warning::
```
corneliusroemer, line 86: This should still be linked somewhere

corneliusroemer (commented):
We should not delete the "how to set osx-64 base env" - this is valuable info for those who don't want to reinstall conda

src/install.rst
victorlin: The option is still available if you click the linked FAQ page which says:

https://github.com/nextstrain/docs.nextstrain.org/blob/5b5f71fcfa65463c3920cfbdd340814d2be9e5d9/src/reference/faq.rst?plain=1#L23-L29

Though it's now simplified to avoid an intermediate `base_osx-64` environment – something I got from https://github.com/conda-forge/miniforge/issues/165#issuecomment-860233092 but only caused more confusion in practice.

victorlin: I see you created #125 for this. Let's continue discussion there.

PR #126, opened 2022-08-10
Add instructions for existing Conda installations

### LINKED ISSUE #125: Add instructions for users who have already installed Conda
In https://github.com/nextstrain/docs.nextstrain.org/pull/124 you seem to have removed the snippets for how to set an Intel emulation environment.

Talking about this snippet:
https://github.com/nextstrain/docs.nextstrain.org/blob/877dbaac5362fa60aff8426cc49962f0b6009888/src/install.rst#L86-L100

This is sad. I've already got the ARM miniconda installed. What do I do now?

I don't think everyone will want to uninstall conda just to install the Nextstrain environment.

We should definitely allow both options. If you set up fresh: it's easiest to use Intel miniconda.

If you've already got conda, this is how you enable Intel emulation environment.

@tsibley @victorlin

victorlin:
_[(preview)](https://nextstrain--126.org.readthedocs.build/en/126/install.html)_

It's likely that some users reading the installation docs will already
have Conda installed. The current docs do not take this into account.
Adding a path for these users, especially those using Apple silicon
computers which may complicate the installation process.

### Related issue(s)

- Closes #125

### Testing

- [ ] Get users to run through new docs?

### DIFF
--- a/src/install.rst
+++ b/src/install.rst
@@ -46,7 +46,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
          If you are an experienced user, you can replace ``conda`` with ``pip`` but :doc:`note the extra installation steps for Augur <augur:installation/installation>` and :doc:`install Auspice via npm <auspice:introduction/install>`.
 
-      1. Install Miniconda:
+      1. Install Miniconda if you don't have ``conda`` yet:
 
          .. The installer link is taken from https://docs.conda.io/en/latest/miniconda.html.
 
@@ -89,6 +89,10 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
             .. group-tab:: Native
 
+               .. warning::
+
+                  If you already had Conda installed prior to following these steps **and** are using a newer Mac with an `Apple silicon chip <https://support.apple.com/en-us/HT211814>`_ (e.g. M1), please :ref:`read this FAQ section to see if you should run an extra step before proceeding <check-existing-conda-installation>`.
+
                1. Create a Conda environment named ``nextstrain``:
 
                   .. include:: snippets/conda-create-bash.rst
@@ -120,7 +124,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
          .. note:: You may have to restart your machine when configuring WSL.
 
       2. `Install Docker Desktop with WSL 2 backend <https://docs.docker.com/desktop/windows/wsl/>`_.
-      3. Install Miniconda:
+      3. Install Miniconda if you don't have ``conda`` yet:
 
          a. Go to the `installation page <https://docs.conda.io/en/latest/miniconda.html>`_.
          b. Scroll down to the **Latest Miniconda Installer Links** section and click the Windows platform link relevant to your machine.
@@ -167,7 +171,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
          .. note:: You may have to restart your machine when configuring WSL.
 
       2. Open a WSL terminal by running **wsl** from the Start menu.
-      3. Install Miniconda:
+      3. Install Miniconda if you don't have ``conda`` yet:
 
          .. code-block:: bash
 
@@ -231,7 +235,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
          If you are an experienced user, you can replace ``conda`` with ``pip`` but :doc:`note the extra installation steps for Augur <augur:installation/installation>` and :doc:`install Auspice via npm <auspice:introduction/install>`.
 
-      1. Install Miniconda:
+      1. Install Miniconda if you don't have ``conda`` yet:
 
          .. code-block:: bash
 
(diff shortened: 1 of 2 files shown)


src/reference/faq.rst
```diff
@@ -20,7 +20,37 @@ Why recommend the Intel Miniconda installer for Mac computers with Apple silicon
...
+
+   - ``x86_64``: This means you have an Intel Conda installation. We recommend keeping this installation method and there is no extra step for Nextstrain installation.
+   - ``arm64``: This means you have an Apple silicon Conda installation. You can choose to keep this installation to retain existing Conda environments, or to re-install with the Intel installer (note that you will have to re-create any existing environments). If you choose to keep this installation, please follow directions in the next section to properly install Nextstrain.
+
+.. _install-nextstrain-on-apple-silicon-conda:
+
+Workaround for installing Nextstrain on an Apple silicon Conda installation
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+For those who really want to use an Apple silicon (``arm64``) Conda installation, it is still possible to set up the ``nextstrain`` Conda environment by configuring it to run with emulation. Run this after setting up the empty Conda environment and before the ``mamba install`` command:
```
huddlej, line 53: > Run this after setting up the empty Conda environment...

I think this statement refers to the empty `nextstrain` environment that the user will create back in the installation instructions, but in the context of this section of the FAQ page, it's possible the user could interpret this as the `tmp` environment they were asked to create above. Could we clarify which empty environment in the last sentence here?

src/reference/faq.rst
```diff
@@ -20,7 +20,37 @@ Why recommend the Intel Miniconda installer for Mac computers with Apple silicon
...
+
+   .. code-block:: bash
+
+      python -c "import platform;print(platform.machine())"
+
+   The output should be one of the following:
+
+   - ``x86_64``: This means you have an Intel Conda installation. We recommend keeping this installation method and there is no extra step for Nextstrain installation.
+   - ``arm64``: This means you have an Apple silicon Conda installation. You can choose to keep this installation to retain existing Conda environments, or to re-install with the Intel installer (note that you will have to re-create any existing environments). If you choose to keep this installation, please follow directions in the next section to properly install Nextstrain.
+
```
huddlej, line 47: Can we include a clean up step here to remove the `tmp` environment that the user just created?

src/reference/faq.rst
```diff
@@ -20,7 +20,37 @@ Why recommend the Intel Miniconda installer for Mac computers with Apple silicon
...
+2. Check output of the following:
+
+   .. code-block:: bash
+
+      python -c "import platform;print(platform.machine())"
+
+   The output should be one of the following:
+
+   - ``x86_64``: This means you have an Intel Conda installation. We recommend keeping this installation method and there is no extra step for Nextstrain installation.
+   - ``arm64``: This means you have an Apple silicon Conda installation. You can choose to keep this installation to retain existing Conda environments, or to re-install with the Intel installer (note that you will have to re-create any existing environments). If you choose to keep this installation, please follow directions in the next section to properly install Nextstrain.
```
huddlej, line 46: I like how clearly you've laid out the options here. It's a bummer, though, that users have to choose between keeping all of their current Conda environments and installing Nextstrain though. I could see some people just stopping here, but maybe not...

huddlej (changes requested):
Thank you for thinking of the users with this use case, @victorlin! This generally looks good. I only have a couple minor requests in the FAQ section.

src/reference/faq.rst
victorlin: Done: 3a1557d2343f926305829c0c35e8afcca76bd261

src/reference/faq.rst
victorlin: Done: ef7c095a09b9f8b887b27e369bc40ed6e55ca2bc

src/reference/faq.rst
victorlin: > users have to choose between keeping all of their current Conda environments and installing Nextstrain

Well not exactly, they can have both by following the directions in the next section. Maybe this can be made more clear?

src/install.rst
```diff
@@ -46,7 +46,7 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
          If you are an experienced user, you can replace ``conda`` with ``pip`` but :doc:`note the extra installation steps for Augur <augur:installation/installation>` and :doc:`install Auspice via npm <auspice:introduction/install>`.
 
-      1. Install Miniconda:
+      1. Install Miniconda if you don't have ``conda`` yet:
```
victorlin, line 49: A user from Bloom lab just messaged me with some weird behavior where Conda environments weren't being created properly. Turns out they had installed Miniconda on top of an existing Anaconda installation, which then caused weird behaviors. This step should explicitly have users check if there is an existing installation by running `conda` in the command line.

victorlin:
Closing in favor of 44c4024402ded3d8bad6f6f07a02e75b7a03144a from #157.

PR #128, opened 2022-08-15
Add hint to create Windows homedir symlink in WSL

### LINKED ISSUE #114: Add note about how to symlink to Windows storage from Linux in WSL installation guide
Following up on [a year-old Slack conversation](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1624641884179500), we should provide instructions for users who run Linux in WSL on Windows to link their Windows storage from Linux in a standard way. This approach allows users to use the Windows explorer interface to manage files instead of the Linux filesystem. A generic solution might be to map the top-level user directory from Windows to a standard name in the Linux filesystem like:

```bash
ln -s /mnt/c/Users/<user> ~/windows_user
```

We should clarify that the `<user>` that users enter here is their _Windows_ username and not the Linux username, since these names can be different.

victorlin:

This will help users who may be confused on how to access Windows files
from WSL.

### Related issue(s)

Closes #114.

### Testing

_N/A_

### DIFF
--- a/src/install.rst
+++ b/src/install.rst
@@ -225,6 +225,19 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
             Setting default environment to <option>.
 
+      .. hint::
+
+         By default, your Windows home directory will not be directly accessible under your WSL home directory. When run in a WSL prompt, the following command fixes that by creating a symlink to it in your WSL home directory. This allows you to use Windows-based text editors and Linux commands all on the same files. Before running,
+
+         1. Replace ``<user>`` with your Windows username.
+         2. Optionally, change the ``windows_home`` folder name or only link to a specific directory under your windows user (e.g. ``ln -s /mnt/c/Users/<user>/Documents ~/windows_documents``).
+
+         .. code-block:: bash
+
+               ln -ws /mnt/c/Users/<user> ~/windows_home
+
+         If your computer's language is set to something other than English, be sure to change the directory names in the path to correspond. For example, on a German-language machine, the path may be ``/mnt/c/Benutzer/<user>``
+
    .. group-tab:: Ubuntu Linux
 
       .. note::


emmahodcroft:
I think this looks good - I do think adding such a shortcut is really, really helpful to Windows users. My only suggestion is to add something like, "If your computer's language is set to something other than English, be sure to change the folder names in the path to correspond. For example, on a German-language machine, the path may be `/mnt/c/Benutzer/<user>`."

I've run into this a number of times on international tutorials!

victorlin:
@emmahodcroft good point, I forgot about that discussion in the linked issue. Added your suggestion, and also:

1. Renamed `windows_user` to `windows_home`.
2. Use the `-w` option to warn if the source directory does not exist (no simple way to error, so the link would still be created but a warning should help).

j23414 (approved):
Happy to see the symlink! :D

emmahodcroft:
Looks great now @victorlin , thanks for including that!

src/install.rst
```diff
@@ -225,6 +225,19 @@ These instructions will install the Nextstrain CLI and tools to run and view you
...
+      .. hint::
+
+         By default, your Windows home directory will not be directly accessible under your WSL home directory. When run in a WSL prompt, the following command fixes that by creating a symlink to it in your WSL home directory. This allows you to use Windows-based text editors and Linux commands all on the same files. Before running,
+
+         1. Replace ``<user>`` with your Windows username.
+         2. Optionally, change the ``windows_home`` folder name or only link to a specific directory under your windows user (e.g. ``ln -s /mnt/c/Users/<user>/Documents ~/windows_documents``).
+
+         .. code-block:: bash
+
+               ln -ws /mnt/c/Users/<user> ~/windows_home
```
tsibley, line 237: Does WSL expose the typical Windows environment variables like `USERPROFILE` or `HOMEDRIVE` and `HOMEPATH`? If so, those could be used in this command to avoid the manual "replace X with Y" steps.

victorlin: Oh good thinking! [From StackExchange](https://superuser.com/a/1568668):

```sh
wslpath "$(wslvar USERPROFILE)"
```

I'll make another PR using this.

PR #129, opened 2022-08-23
Simplify WSL symlink recommendation

victorlin:

Use an existing variable to get the link's source path.

### Related issue(s)

- Related to https://github.com/nextstrain/docs.nextstrain.org/pull/128#discussion_r951947234

### Testing

- [x] Manually test that the new command works on an English Windows machine.
- [x] Manually test that the new command works on a non-English Windows machine: https://github.com/nextstrain/docs.nextstrain.org/pull/129#discussion_r953636956

### DIFF
--- a/src/install.rst
+++ b/src/install.rst
@@ -227,16 +227,15 @@ These instructions will install the Nextstrain CLI and tools to run and view you
 
       .. hint::
 
-         By default, your Windows home directory will not be directly accessible under your WSL home directory. When run in a WSL prompt, the following command fixes that by creating a symlink to it in your WSL home directory. This allows you to use Windows-based text editors and Linux commands all on the same files. Before running,
-
-         1. Replace ``<user>`` with your Windows username.
-         2. Optionally, change the ``windows_home`` folder name or only link to a specific directory under your windows user (e.g. ``ln -s /mnt/c/Users/<user>/Documents ~/windows_documents``).
+         By default, your Windows home directory will not be directly accessible under your WSL home directory. When run in a WSL prompt, the following command fixes that by creating a symlink to it in your WSL home directory. This allows you to use Windows-based text editors and Linux commands all on the same files.
 
          .. code-block:: bash
 
-               ln -ws /mnt/c/Users/<user> ~/windows_home
+               ln -ws "$(wslpath "$(wslvar USERPROFILE)")" ~/windows_home
 
-         If your computer's language is set to something other than English, be sure to change the directory names in the path to correspond. For example, on a German-language machine, the path may be ``/mnt/c/Benutzer/<user>``
+         Optionally, you can customize the ``windows_home`` folder name or only link to a specific directory under your windows user (e.g. ``ln -ws "$(wslpath "$(wslvar USERPROFILE)")/Documents" ~/windows_documents``).
+
+         If the command does not work, you may have to first run ``sudo apt install wslu``.
 
    .. group-tab:: Ubuntu Linux
 


src/install.rst
```diff
@@ -227,16 +227,15 @@ These instructions will install the Nextstrain CLI and tools to run and view you
...
 
          .. code-block:: bash
 
-               ln -ws /mnt/c/Users/<user> ~/windows_home
+               ln -ws $(wslpath $(wslvar USERPROFILE)) ~/windows_home
+
+         Optionally, change the ``windows_home`` folder name or only link to a specific directory under your windows user (e.g. ``ln -s $(wslpath $(wslvar USERPROFILE))/Documents ~/windows_documents``).
 
-         If your computer's language is set to something other than English, be sure to change the directory names in the path to correspond. For example, on a German-language machine, the path may be ``/mnt/c/Benutzer/<user>``
+         If the command does not work, you may have to first run `sudo apt install wslu`.
```
victorlin, line 238: @emmahodcroft do you have a computer that was set up in a language other than English? If so, could you verify that the new command (or even just `ls $(wslpath $(wslvar USERPROFILE))`) works? I assume it will since this is a variable from the system itself, but I want to be sure before removing the caveat.

I tried setting my system language to Chinese but the paths are still in English.

The only info I could find on this was [this claim from a Microsoft discussion post](https://answers.microsoft.com/en-us/windows/forum/all/rename-system-folders-in-another-language/be6b6b94-d5ea-48af-92a0-006930070ce3):

> The base folder names are always in English. You can just type the name and it will work.

emmahodcroft: Yea, I think if you change a language later, it keeps all the system paths (understandable I suppose, as otherwise things would be very broken).

Unfortunately I'm the only one connected to Richard's group that uses Windows... let me reach out a little more widely and see if I can find a guinea pig.

src/install.rst
```diff
@@ -227,16 +227,15 @@ These instructions will install the Nextstrain CLI and tools to run and view you
...
-         1. Replace ``<user>`` with your Windows username.
-         2. Optionally, change the ``windows_home`` folder name or only link to a specific directory under your windows user (e.g. ``ln -s /mnt/c/Users/<user>/Documents ~/windows_documents``).
+         By default, your Windows home directory will not be directly accessible under your WSL home directory. When run in a WSL prompt, the following command fixes that by creating a symlink to it in your WSL home directory. This allows you to use Windows-based text editors and Linux commands all on the same files.
 
          .. code-block:: bash
 
-               ln -ws /mnt/c/Users/<user> ~/windows_home
+               ln -ws $(wslpath $(wslvar USERPROFILE)) ~/windows_home
+
+         Optionally, change the ``windows_home`` folder name or only link to a specific directory under your windows user (e.g. ``ln -s $(wslpath $(wslvar USERPROFILE))/Documents ~/windows_documents``).
```
emmahodcroft, line 236:
```suggestion
         Optionally, you can customize the ``windows_home`` folder name or only link to a specific directory under your windows user (e.g. ``ln -s $(wslpath $(wslvar USERPROFILE))/Documents ~/windows_documents``).
```
Just thought this wording is slightly clearer.

src/install.rst
```diff
@@ -227,16 +227,15 @@ These instructions will install the Nextstrain CLI and tools to run and view you
...
-         By default, your Windows home directory will not be directly accessible under your WSL home directory. When run in a WSL prompt, the following command fixes that by creating a symlink to it in your WSL home directory. This allows you to use Windows-based text editors and Linux commands all on the same files. Before running,
-
-         1. Replace ``<user>`` with your Windows username.
-         2. Optionally, change the ``windows_home`` folder name or only link to a specific directory under your windows user (e.g. ``ln -s /mnt/c/Users/<user>/Documents ~/windows_documents``).
+         By default, your Windows home directory will not be directly accessible under your WSL home directory. When run in a WSL prompt, the following command fixes that by creating a symlink to it in your WSL home directory. This allows you to use Windows-based text editors and Linux commands all on the same files.
 
          .. code-block:: bash
 
-               ln -ws /mnt/c/Users/<user> ~/windows_home
+               ln -ws $(wslpath $(wslvar USERPROFILE)) ~/windows_home
```
tsibley, line 234: This needs quotes to be safe. 

```suggestion
               ln -ws "$(wslpath "$(wslvar USERPROFILE)")" ~/windows_home
```

tsibley (changes requested):

src/install.rst
victorlin: Done: c7d27becfddafa7f4ddb1fb75d86af715bb1de8b

src/install.rst
victorlin: Done: 171da63d1032010d5e6a2b554f5bfbb63833c36a

src/install.rst
victorlin: Great, thanks for reaching out!

emmahodcroft: So, I don't know what to make of this - their directories in Windows explorer are in Dutch, but when they use the command, it returns `/mnt/c/Users/<username>/` not `C:\Gebruikers\<username>` as it shows in Explorer. When they check the 'properties' (I guess they mean folder properties?) it's listed as `C:\Users\<username>`.

This would seem to imply that underlying path can be in English and just displayed in the interface in the language of the system - but in that case, I'm confused how I ran into this problem when we did the tutorial at RIVM (Netherlands) - as our English-language command should have worked fine if that's true, but didn't, and had to be adjusted to other EU languages... 

But I don't know how likely it is we get a large test audience for this - there aren't so many people out there with non-English-language computers and also WSL, it seems. We could just change to this command, and I guess we'll get issues/complaints/find out in a tutorial if it doesn't work as expected...?

src/install.rst
victorlin: Forgot to apply this to the alternative example. Done: 9e88c29d60c80006efc3da643e3e762a5aceed1a

src/install.rst
victorlin: That behavior is in line with the claim I mentioned above, and the same as what I experienced when changing English to Chinese ("The base folder names are always in English").

I'll just merge this as-is and hopefully any issues are brought to our attention.

For your issues at RIVM, if it was a while ago then maybe people were using WSL 1 instead of WSL 2? There might be some differences in how the C drive gets mounted between the two versions.

emmahodcroft: Fair point - this was very likely with WSL1, and they may have made the C drive mounting more redundant to avoid these issues. Anyway, presumably if it is an issue we'll run into it eventually and we can address it with a similar comment as we had previously if we need to, about being aware of non-English-language systems.

victorlin:
Merging without explicit approval since I think there's been enough scrutiny on this, and it's easy to revert/change.

PR #130, opened 2022-08-26
dev: Add "livehtml" make target using sphinx-autobuild

tsibley:
Super handy for edit/preview cycles in dev:

    make livehtml

Then open <http://localhost:8000>.

### Testing
- [x] Works locally.
- [x] Doesn't loop forever!

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,6 +19,15 @@ help:
 %: Makefile
 	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
+# Generating --ignore options from .gitignore means that src/fetch-docs.py,
+# which runs on build and produces files in src/, doesn't trigger an endless
+# loop of autobuilds.  The syntax of .gitignore patterns is not actually
+# identical to sphinx-autobuild's --ignore patterns (which use Python's
+# fnmatch(), where "*" means something other than a normal glob, for example),
+# but our usage is limited enough for this to work as much as necessary.
+livehtml:
+	sphinx-autobuild -b html $(patsubst %,--ignore "*/%",$(file < .gitignore)) "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
+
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -19,6 +19,14 @@ help:
 %: Makefile
 	@$(SPHINXBUILD) -M $@ "$(SOURCEDIR)" "$(BUILDDIR)" $(SPHINXOPTS) $(O)
 
+# Generating --ignore options from .gitignore means that src/fetch-docs.py,
+# which runs on build and produces files in src/, doesn't trigger an endless
+# loop of autobuilds.  The full syntax of .gitignore patterns is not supported
+# by sphinx-autobuild's --ignore globs, but our usage is limited enough for
+# this to work as much as necessary.
```
victorlin, line 26: Hmm, I'm still getting an infinite loop due to detected changes on `src/guides/communicate/narratives-intro.md`.

Commenting out these lines makes it work fine (this is also what I've been doing):

https://github.com/nextstrain/docs.nextstrain.org/blob/3398c8b97bf4d1ddd09db716e663bd8e1007de3b/src/conf.py#L19-L20

tsibley: What's the first few lines of output of `make livehtml` for you?

tsibley: And what's the output of `printf '%s\n' "$PWD" "$(realpath "$PWD")"` where ever you're running `make livehtml`? My suspicion is that the `$(PWD)/` prefixing in the Makefile isn't matching and that '*/` prefixing might be better anyway.

victorlin: I don't have `realpath` installed, but hopefully the info above is enough to understand the problem.

tsibley: Got it in one. \o/

The problem is as I suspected. Should be fixed by repush.

tsibley: sphinx-autobuild matches against fully-resolved paths (e.g. what `realpath` produces), so we need to produce those (what I was doing with `$(PWD)`) or explicitly match a suffix (what I'm now doing with `*/`).

victorlin: Works great now, thanks!

PR #131, opened 2022-10-03
Update runtime names

victorlin:

In anticipation of upcoming changes to the Nextstrain CLI [[1]].

[1]: https://github.com/nextstrain/cli/pull/218

### Related issue(s)

_N/A_

### Tasks

- [x] CLI: merge https://github.com/nextstrain/cli/pull/218
- [x] CLI: Rename `native` to `ambient`
- [x] CLI: release
- [x] Update CLI version placeholders `x.x.x` in this PR

### Testing

- [x] Checks pass
- [x] Manually checked preview page

### DIFF
--- a/src/install.rst
+++ b/src/install.rst
@@ -22,6 +22,7 @@ For help choosing, refer to our :doc:`/reference/faq`, such as:
 
   * :ref:`what-are-docker-conda-mamba-wsl-etc`
   * :ref:`choosing-a-runtime`
+  * :ref:`what-happened-to-the-native-runtime`
   * :ref:`when-to-use-wsl`
 
 First, install a Nextstrain runtime.
--- a/src/install.rst
+++ b/src/install.rst
@@ -17,7 +17,7 @@ When completed, you'll be ready to run Nextstrain :term:`workflows <workflow>`.
 Installation steps
 ==================
 
-Steps vary by runtime option (Docker, native) and operating system (macOS, Windows, WSL on Windows, Linux).
+Steps vary by runtime option (Docker, ambient) and operating system (macOS, Windows, WSL on Windows, Linux).
 For help choosing, refer to our :doc:`/reference/faq`, such as:
 
   * :ref:`what-are-docker-conda-mamba-wsl-etc`
@@ -36,7 +36,7 @@ First, install a Nextstrain runtime.
 
             .. warning::
 
-               If using a newer Mac with an `Apple silicon chip <https://support.apple.com/en-us/HT211814>`_ (e.g. M1), the **native** runtime is recommended due to slowness with the Docker runtime. `We are considering ways to improve this <https://github.com/nextstrain/docker-base/issues/35>`_.
+               If using a newer Mac with an `Apple silicon chip <https://support.apple.com/en-us/HT211814>`_ (e.g. M1), the **ambient** runtime is recommended due to slowness with the Docker runtime. `We are considering ways to improve this <https://github.com/nextstrain/docker-base/issues/35>`_.
 
             1. `Install Docker Desktop using the official guide <https://docs.docker.com/desktop/install/mac-install/>`_.
             2. Install the Nextstrain CLI:
@@ -115,7 +115,7 @@ First, install a Nextstrain runtime.
             5. Follow the installer's final instructions to setup your shell config.
 
 
-   .. group-tab:: Native
+   .. group-tab:: Ambient
 
       .. tabs::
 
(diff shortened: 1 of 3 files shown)


tsibley:
Repush to replace `x.x.x` with `5.0.0` and fix nested formatting that rST (incomprehensibly) doesn't support:

```rst
.. inline formatting not nestable
**new ``conda`` runtime that runs natively**

.. so I did this
**new** ``conda`` **runtime that runs natively**
```

tsibley:
This PR also needs an update to switch `native` to `ambient` on the install page, though it doesn't prevent anything from working if folks use `native`.

tsibley:
@victorlin I've added some commits to this PR to round out the update. Take a look and if you agree it's good enough as-is for now, let's merge?

victorlin:
@tsibley thanks for doing that! I added two more fixups: ea67b3e...229cefd

Will squash those in and merge.

src/reference/faq.rst
```diff
@@ -83,3 +85,34 @@ For those who really want to use the ``arm64`` Miniconda installer, it is still
...
+.. _what-happened-to-the-native-runtime:
+
+What happened to the ``native`` runtime?
+----------------------------------------
+
+Up until Nextstrain CLI version 5.0.0, the name of the ``native`` runtime has been debated internally. In CLI version 5.0.0, **it was renamed to** ``ambient`` and we will use the new name going forwards.
+
+"Native" as a software term is typically used to describe software that can run without emulation, in other words optimized for your computer's processor.
+
+The ``ambient`` is "native" in that sense, but it puts all the software maintenance burden on the user. This means:
```
victorlin:
```suggestion
The ``ambient`` runtime is "native" in that sense, but it puts all the software maintenance burden on the user. This means:
```

victorlin: Done in [force-push](https://github.com/nextstrain/docs.nextstrain.org/compare/6defc78df4d90e9e7e2a3d3bb2a8f4cd00cfe0fa..e731803c2442314f9ecc0a7f0d2b2569adb5daa0).

src/reference/faq.rst
```diff
@@ -83,3 +85,34 @@ For those who really want to use the ``arm64`` Miniconda installer, it is still
...
+2. It is up to you as the creator of the ``nextstrain`` Conda environment to know (1) how to activate it, (2) when to update it, and (3) how to update it.
+
+So really, the ``ambient`` runtime is any environment that has been set up with all of the required software available on your local ``PATH``. We chose Conda in the installation instructions since some users may already be familiar with it, and it is simpler than using individual package managers for the various required software (e.g. ``pip``, ``npm``).
+
+Most importantly, Nextstrain CLI version 5.0.0 provides a **new** ``conda`` **runtime that runs natively** without putting all of the software maintenance burden on users. This means the ``ambient`` runtime is no longer the only "native" runtime, and we will recommend new users to use the ``conda`` runtime instead of ``ambient``.
+
+``ambient`` is still a good option for users who wish to customize their environment to include other software used in their workflows.
```
victorlin, line 107: I don't think this PR is ready, since the wording assumes that Conda runtime installation instructions are available when they aren't yet.

src/install.rst
```diff
@@ -17,11 +17,12 @@ When completed, you'll be ready to run Nextstrain :term:`workflows <workflow>`.
 Installation steps
 ==================
 
-Steps vary by runtime option (Docker, native) and operating system (macOS, Windows, WSL on Windows, Linux).
+Steps vary by runtime option (Docker, ambient) and operating system (macOS, Windows, WSL on Windows, Linux).
```
victorlin, line 20: Re-ordered the commits so that 5bc7a5d comes before 18a6a0f, since 18a6a0f references the new name.

tsibley: Thanks!

tsibley:
> I added two more fixups: [ea67b3e...229cefd](https://github.com/nextstrain/docs.nextstrain.org/compare/ea67b3e17b192aa47dd14f9fb38a346ad7731c25...229cefd68ac39d56cb35099119b1fd31e9f93c02)

:facepalm: I missed those because I skipped over `faq.rst` thinking the usages were all from the new FAQ entries. orz

Thanks for catching them!

src/reference/faq.rst
tsibley: Ah, I didn't read it that way, though I see what you mean.

> and we will recommend new users to use the ``conda`` runtime instead of ``ambient``.

I think this is true (i.e. we _will_ do it in the future), even though we're not quite at that point yet.

It seems ok to merge with this to me, but if you have revised wording you'd like, happy to go with that too.

victorlin: Oh I see, yeah this should be fine then. I'll merge.

PR #110, opened 2022-05-24
groups: Revise timeline since we didn't start as quickly as I hoped

tsibley:
Dates still to be determined for now pending internal migrations first
and some questions around the CLI behaviour.

### Related issue(s)
Related to #104 

### Testing
- [x] Looks good locally
- [x] Check CI and preview build

PR #137, opened 2022-10-27
Kick off groups migration

tsibley:
New proposed timeline now that we've agreed on Nextstrain CLI behaviour and Nextstrain CLI 5.0.0 is out with the `nextstrain remote download` behaviour change (+ groups-related bug fixes, `--dry-run` features, and more).

Timeline is mostly arbitrary beyond "sooner than later", so please consider it a strawman if you have suggestions!

Related to #104
Related to #110

### DIFF
--- a/src/guides/share/groups/migrate-from-s3.rst
+++ b/src/guides/share/groups/migrate-from-s3.rst
@@ -168,19 +168,17 @@ Timeline
 :March 2022: New groups created after this point manage their data through
              nextstrain.org instead of S3.
 
-:TBD: `Notification`_ sent to groups created prior to March.
+:early November: `Notification`_ sent to groups created prior to March.
 
-:TBD: Groups migrated one-by-one in coordination with group owners.
+:November onwards: Groups start migrating one-by-one in coordination with group owners.
 
-:TBD: Nextstrain team's desired deadline for migrating all groups.
+:end of February 2023: Nextstrain team's desired deadline for migrating all groups.
 
 
 Notification
 ============
 
-.. note:: Notifications have not yet been sent out.
-
-In *[TBD]* 2022, we emailed the contacts for all groups created before March 2022
+In November 2022, we emailed the contacts for all groups created before March 2022
 to inform them of these changes.  A copy of the email is below.
 
     **Subject:** Action requested: Improvements to how you manage your
--- a/src/guides/share/groups/migrate-from-s3.rst
+++ b/src/guides/share/groups/migrate-from-s3.rst
@@ -23,16 +23,16 @@ dataset to a Nextstrain Group:
     # Old way
     nextstrain remote upload \
       s3://nextstrain-${GROUPNAME}/ \
-        auspice/ncov_${YOUR_BUILD_NAME}.json \
-        auspice/ncov_${YOUR_BUILD_NAME}_tip-frequencies.json \
-        auspice/ncov_${YOUR_BUILD_NAME}_root-sequence.json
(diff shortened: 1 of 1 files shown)


j23414 (approved):
As far as I can understand, the timeline makes sense.

tsibley:
@j23414 :+1: let me know if I can clarify!

I'm going to tag @nextstrain/core another time or two to try to get a few more responses before I blast out emails to users.

joverlee521 (approved):
Timelines make sense!

Should we explicitly mention that after migration they will lose access to the old S3 bucket and that the bucket will be completely deleted after a grace period?

tsibley:
> Should we explicitly mention that after migration they will lose access to the old S3 bucket and that the bucket will be completely deleted after a grace period?

Good point! Noted in f46ca2b. I've also updated https://github.com/nextstrain/private/issues/50 to note the 30 day minimum promised.

The migration process does alert the operator (i.e. me, most likely) when there are non-dataset, non-narrative files that need manual review. My plan in most cases would be to alert the group owners and ask if they'd like copies of them.

src/guides/share/groups/migrate-from-s3.rst
```diff
@@ -168,19 +169,17 @@ Timeline
...
 Notification
 ============
 
-.. note:: Notifications have not yet been sent out.
-
-In *[TBD]* 2022, we emailed the contacts for all groups created before March 2022
+In November 2022, we emailed the contacts for all groups created before March 2022
 to inform them of these changes.  A copy of the email is below.
 
     **Subject:** Action requested: Improvements to how you manage your
```
victorlin, line 185: To me, the email doesn't convey the "Action requested" clearly – one has to click on the link and get through the What/Why sections to finally see the numbered list of requests.

I'd suggest adding a short sentence or two describing the migration, followed by the numbered list of requests without the details. This can then be followed by a link to the docs page to read more about it.

victorlin (approved):
Changes look good 👍

The email contents aren't a part of this PR, but I just read through it for the first time and have some suggestions.

src/guides/share/groups/migrate-from-s3.rst
tsibley: Ah, so I intentionally omitted the `nextstrain remote` changes and an actual list of steps from the email. The idea was that I don't want folks to only read a summarized version in the email and not click thru because that will a) leave them without enough info and b) it isn't possible to update the email on-the-fly if we want to make clarifications as some migrations get started but others remain pending. (The "action requested" in the subject line was to indicate "This email requires a response", not so much as to describe the contents.)

Thoughts on balancing being too vague with trying to make sure folks click thru?

Prompted by your comment, I thought it good to shift things around ([3c14509](https://github.com/nextstrain/docs.nextstrain.org/pull/137/commits/3c14509e20548c3f00b91aae3345e4942d826a0b):

image

But this still doesn't address your concern of not including an actual action item list.

victorlin: That looks good! My only concern for vagueness is lack of response. For those who don't respond the first time, we can send a reminder email that is more verbose.

victorlin (approved):

src/guides/share/groups/migrate-from-s3.rst
tsibley: Ok, sounds good. Yeah, I expect we'll have to send some reminders regardless of what we do, but we'll see! :-)

PR #138, opened 2022-10-28
install: Add instructions for the Conda runtime

tsibley:
See commit messages for details and read [the preview](https://nextstrain--138.org.readthedocs.build/en/138/install.html).

Related to https://github.com/nextstrain/cli/pull/218 and others in the CLI repo.
Related to #131.

### Testing

- [x] Checks pass
- [x] Preview looks good

### DIFF
--- a/src/install.rst
+++ b/src/install.rst
@@ -149,16 +149,25 @@ Then, install a Nextstrain runtime.
 
    .. group-tab:: Conda
 
-      .. note::
+      .. tabs::
 
-         Due to installation constraints, there is no way to use Nextstrain's Conda runtime on Windows directly. Starting from the beginning, follow steps for **WSL on Windows** if the Conda runtime is desired, or use the **Docker** runtime instead.
+         .. group-tab:: macOS
 
-      Set up the runtime:
+            .. include:: snippets/nextstrain-setup-conda.rst
 
-      .. code-block:: none
+         .. group-tab:: Windows
 
-         nextstrain setup --set-default conda
+            .. note::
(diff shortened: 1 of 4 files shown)


src/install.rst
```diff
@@ -17,225 +17,280 @@ When completed, you'll be ready to run Nextstrain :term:`workflows <workflow>`.
...
+         .. code-block:: none
+
+            nextstrain setup --set-default docker
+
+
+   .. group-tab:: Conda
+
+      .. note::
+
+         Due to installation constraints, there is no way to use Nextstrain's Conda runtime on Windows directly. Starting from the beginning, follow steps for **WSL on Windows** if the Conda runtime is desired, or use the **Docker** runtime instead.
```
victorlin, line 154: Non-Windows users might be confused by this message. I think it's unnecessary to show for those users, and adding tabs here to be consistent with other runtime tabs would make for a better user experience. Example: 2c5f4d612b0d57097bda0b03487aaba07ec0c7da

victorlin (changes requested):

src/install.rst
tsibley: Good point. I avoided tabs here to try and simplify the appearance, but looking at your example, I agree it's better. Applied 2c5f4d612b0d57097bda0b03487aaba07ec0c7da as c2e954c977df72fffbd3fbf4c57eb4c50778cb74 after fixing indentation and adding your comment above as the commit message.

victorlin (approved):

PR #145, opened 2022-11-19
Rename RTD config file

victorlin:

From the official doc¹:

> The configuration file should be named .readthedocs.yaml.
> Some other variants … are deprecated.

¹ https://docs.readthedocs.io/en/stable/config-file/index.html

Based on a [GitHub search](https://github.com/search?q=org%3Anextstrain+path%3A*readthedocs.yml&type=code), this applies to 8 other repos under @nextstrain.

Even though the current name works for now, the wording in the official doc makes it unreliable.

### Related issue(s)

_N/A_

### Testing

- [ ] Checks pass
- [ ] Preview build works


tsibley (commented):
I don't really think this is at all necessary. While RTD says the other name variants are deprecated, I really don't think they're going to remove support for them any time soon (if ever). There are [over 30k instances](https://github.com/search?l=&q=filename%3Areadthedocs.yml&type=code) of `readthedocs.yml` or `.readthedocs.yml` (both deprecated) on GitHub alone.

It'd be fine to make this change, but it feels like needless shuffling, esp. if the idea is to go make similar renames in several other repos too.

Separately, I do think it's nice to not have the RTD config file hidden away as a dotfile since this whole repo's primary purpose is as an RTD project (unlike our other repos, where it is typically a dotfile).

victorlin:
Valid, will leave things as-is.

PR #147, opened 2022-12-13
install: Document that Rosetta 2 must be enabled for arm64 Macs and how to do so

tsibley:
Related to [this Slack thread](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1670963819691499).

### Testing

- [x] Command allowed standalone Nextstrain CLI setup to succeed
- [x] Checks pass

### DIFF
--- a/src/install.rst
+++ b/src/install.rst
@@ -39,6 +39,18 @@ First, install Nextstrain CLI.
 
       You can launch a Terminal by clicking the Launchpad icon in the Dock, typing ``terminal`` in the search field, and clicking Terminal.
 
+      .. warning::
+
+         If the installation errors with the message "**Bad CPU type in executable**", then you're likely using a newer Mac with an `Apple silicon chip <https://support.apple.com/en-us/HT211814>`_ (e.g. M1).
+
+         `Enable Rosetta 2 <https://support.apple.com/en-us/HT211861>`__ with:
+
+         .. code-block:: bash
+
+            softwareupdate --install-rosetta
+
+         and then retry the installation.
+         Rosetta 2 is required for both Nextstrain CLI itself and our runtimes.
 
    .. group-tab:: Windows
 


tsibley:
I don't like the big warning this leads off our install docs with.

image

but we should put this information somewhere, and ideally _before_ a user runs into the `Bad CPU type in executable` issue. Maybe less ideally would be shortly after:

image

But I wonder about instead updating the installer program to detect the need for Rosetta 2 and error out earlier with the instructions instead of continuing and then failing later.

We should also build an arm64 binary for macOS, but (until now) haven't had the machine to do it as part of CI.

I'm not sure a native binary alone is sufficient, however, since our runtimes will still need Rosetta enabled and command-line invocations don't seem to trigger the prompt.

jameshadfield:
2nd screenshot looks great!

victorlin:
Another vote for the 2nd option. I suspect that this warning won't apply to most users (i.e. few users will come across Nextstrain installation without Rosetta). Of the applicable users, the remedy for an error with cryptic text "Bad CPU type in executable" should be fairly search-able once this text is on the install doc.

> But I wonder about instead updating the installer program to detect the need for Rosetta 2 and error out earlier with the instructions instead of continuing and then failing later.

This sounds better to me, as long as it isn't difficult to implement/maintain.

corneliusroemer:
We could make installing rosetta2 just a normal step of installation:

- Check if rosetta2 is enabled:
```
/usr/bin/pgrep -q oahd && echo "" || echo "Please install Rosetta2 using `softwareupdate --install rosetta`" before proceeding
```

tsibley:
@jameshadfield @victorlin Ok, check out the latest push. I tweaked wording a bit.

image

> This sounds better to me, as long as it isn't difficult to implement/maintain.

It shouldn't be, but I don't know for sure until I try. Also, feel free to give it a try too!

It'd also be nice to detect in the installer that `nextstrain` failed to run. Currently the error message is a bit buried and the installer thinks it succeeded, but when you run the final setup commands it suggests, you also get an error (and will on every subsequent shell init).

corneliusroemer:
Maybe something like this?

```
/usr/bin/pgrep -q oahd && echo "Rosetta already installed" || softwareupdate --install rosetta       
```

That would take care of it automatically

tsibley:
@corneliusroemer Yes, but Rosetta 2 isn't available on Intel Macs, so we'd have to check that too. I also briefly considered splitting our macOS tab into "macOS (Intel)" and "macOS (Apple Silicon)" but disregarded it as requiring the user to know too much when we could detect it.


>     /usr/bin/pgrep -q oahd && echo "" || echo "Please install Rosetta2 using `softwareupdate --install rosetta`" before proceeding

Note that this would execute `softwareupdate --install rosetta` in order to interpolate its output into the string to `echo`. :upside_down_face:

tsibley:
> It'd also be nice to detect in the installer that `nextstrain` failed to run. Currently the error message is a bit buried and the installer thinks it succeeded, but when you run the final setup commands it suggests, you also get an error (and will on every subsequent shell init).

Implemented as https://github.com/nextstrain/cli/pull/241

corneliusroemer:
@tsibley You're right - need to check the architecture.

What about:

```bash
if [[ $(uname -m) -eq arm64 ]] then
    /usr/bin/pgrep -q oahd && echo "Rosetta already installed" || softwareupdate --install rosetta       
fi
```

tsibley:
@corneliusroemer 

I'd want to try and make the test more specific than `pgrep oahd`, which seems too short to be specific enough, e.g. maybe:

```
pgrep -qf /usr/libexec/rosetta/oahd
```

instead.

Also, I don't think we should auto-install it for folks, esp. given how hard it is to uninstall.

It seems like we should still document this even if we add detection, though maybe we can tone down the warning.

tsibley:
Merging this now. I'll add a check to the standalone installer and once _that's_ merged, update the warning here to tone it down and be more of a quiet "heads up" instead.

PR #150, opened 2023-01-31
Add "Components" section to our TOC

tsibley:
before / after
   

See commit messages for details.

### Testing

- [x] Preview looks ok
- [x] Checks pass

### DIFF
--- a/src/index.rst
+++ b/src/index.rst
@@ -81,6 +81,7 @@ team and other Nextstrain users provide assistance.  For private inquiries,
     Nextstrain CLI <https://docs.nextstrain.org/projects/cli/>
     Nextclade CLI <https://docs.nextstrain.org/projects/nextclade/page/user/nextclade-cli.html>
     Nextalign CLI <https://docs.nextstrain.org/projects/nextclade/page/user/nextalign-cli.html>
+    nextstrain.org <https://docs.nextstrain.org/projects/nextstrain-dot-org/>
 
 .. toctree::
     :maxdepth: 1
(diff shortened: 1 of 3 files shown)


emmahodcroft:
I like this!

emmahodcroft (approved):

src/index.rst
```diff
@@ -54,10 +54,10 @@ team and other Nextstrain users provide assistance.  For private inquiries,
     Installing <install>
     tutorials/running-a-workflow
     tutorials/creating-a-workflow
-    Exploring SARS-CoV-2 evolution <https://docs.nextstrain.org/projects/ncov/en/latest/index.html>
+    Exploring SARS-CoV-2 evolution <https://docs.nextstrain.org/projects/ncov/page/index.html>
```
victorlin, line 57: Didn't know about the `/page/` redirects. Thanks for sharing it here! This would be good to add to the "How-to tips" section of the README.

Also for the description of c43ef7fd7f8d79fe104c1aaebd88fc3d5eedd56b, I think you meant `…/page/…` instead of `…/pages/…`, but no big deal.

victorlin (approved):

src/index.rst
tsibley: Ah, good catch! Fixed in 8287793:

image


> This would be good to add to the "How-to tips" section of the README.

I agree it'd be good to note this somewhere… but the README feels enough like a sprawling mess (to me) right now that I don't want to touch it. Instead, I've added it to our fledgling style guide in 9fc90c5.

PR #154, opened 2023-03-08
groups/customize: Overhaul page to apply to new groups and include more information

tsibley:
There are many fewer legacy, single-bucket groups than new, multi-tenant bucket groups now.  Document appropriate curl commands, even if they're not ideal.  Better to have instructions to point to than nothing at all.

### Testing

- [x] Local render looks good
- [x] [Preview](https://nextstrain--154.org.readthedocs.build/en/154/guides/share/groups/customize.html) looks good
- [x] Checks pass

### DIFF
--- a/src/guides/share/groups/customize.rst
+++ b/src/guides/share/groups/customize.rst
@@ -24,7 +24,7 @@ You can customize the content of your group's page by uploading two files to the
 
 Create a new file named ``group-overview.md`` that will contain information about your group.
 At the top of this file, provide a title for the page, a list of people who maintain the data, a website, and/or whether to show datasets and narratives from your group.
-This information is technically known as the `YAML front matter <https://jekyllrb.com/docs/front-matter/>`_ for the file.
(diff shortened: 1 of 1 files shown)


src/guides/share/groups/customize.rst
```diff
@@ -24,7 +24,7 @@ You can customize the content of your group's page by uploading two files to the
 
 Create a new file named ``group-overview.md`` that will contain information about your group.
 At the top of this file, provide a title for the page, a list of people who maintain the data, a website, and/or whether to show datasets and narratives from your group.
-This information is technically known as the `YAML front matter <https://jekyllrb.com/docs/front-matter/>`_ for the file.
+This information is technically known as the `YAML front matter <https://jekyllrb.com/docs/front-matter/>`__ for the file.
```
victorlin: I didn't realize there was a significant difference between `_` and `__`. Thanks for pointing it out!

victorlin (approved):

PR #162, opened 2023-07-25
Add Snakemake style guide

victorlin:
_[preview of new page](https://nextstrain--162.org.readthedocs.build/en/162/reference/snakemake-style-guide.html)_

### Description of proposed changes

Publish the internal Snakemake style guide as a page on the public docs.

### Related issue(s)

This was discussed on Slack [[1](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1670353442520569?thread_ts=1670353195.208659&cid=C01LCTT7JNN)] [[2](https://bedfordlab.slack.com/archives/C01LCTT7JNN/p1683187533217929)].

### Testing

- [x] Checks pass
- [ ] Post-merge: redirect sync is successful
- [ ] Post-merge: "redirect" from the [internal page](https://nextstrain.atlassian.net/wiki/spaces/NEXTSTRAIN/pages/141852677/Snakemake+style+guide)https://[nextstrain.atlassian.net/wiki/spaces/NEXTSTRAIN/pages/141852677/Snakemake+style+guide](https://nextstrain.atlassian.net/wiki/spaces/NEXTSTRAIN/pages/141852677/Snakemake+style+guide) to the new page
- [ ] Post-merge: "redirect" from the [blab styleguide](https://github.com/blab/styleguide) to the new page

PR #164, opened 2023-07-25
Migrate blab/styleguide to the Snakemake style guide

victorlin:

This old styleguide contains recommendations that are relevant to the pathogen workflows presently maintained.

### Related issue(s)

Prompted by https://github.com/nextstrain/docs.nextstrain.org/pull/162#pullrequestreview-1546026324.

### Checklist

- [x] Checks pass
- [x] [Preview page](https://nextstrain--164.org.readthedocs.build/en/164/reference/snakemake-style-guide.html) looks good
- [ ] Post-merge: update the [nextstrain-builds.md](https://github.com/blab/styleguide/blob/HEAD/nextstrain-builds.md) document to link to the [new page](https://docs.nextstrain.org/page/reference/snakemake-style-guide.html)
- [ ] Post-merge: update the [README](https://github.com/blab/styleguide/blob/HEAD/README.md) to say this is unused/never used, and archive the repo (so old links still work)

### DIFF
--- a/src/reference/snakemake-style-guide.rst
+++ b/src/reference/snakemake-style-guide.rst
@@ -194,3 +194,7 @@ Always use relative paths
 Relative paths (paths that don't start with ``/``) mean that anyone can
 run the build without running into portability issues caused by paths
 specific to your computer.
+
+See the `Snakemake documentation
+<https://snakemake.readthedocs.io/en/stable/project_info/faq.html#how-does-snakemake-interpret-relative-paths>`__
+for how relative paths are interpreted depending on context.
--- a/src/reference/snakemake-style-guide.rst
+++ b/src/reference/snakemake-style-guide.rst
@@ -43,9 +43,9 @@ Do this instead of using ``rule`` variables.
 Avoid the ``message`` rule attribute
 ====================================
 
--  When the ``message`` attribute is defined, Snakemake suppresses other
-   critical details that otherwise get displayed by default (e.g., job
-   id, rule name, input, output, etc.).
+When the ``message`` attribute is defined, Snakemake suppresses other critical
+details that otherwise get displayed by default (e.g., job id, rule name,
+input, output, etc.).
 
 Access ``config`` values appropriately
 ======================================
--- a/src/reference/snakemake-style-guide.rst
+++ b/src/reference/snakemake-style-guide.rst
@@ -75,30 +75,20 @@ Example:
    params:
        key=lambda w: config["value_may_contain_curlies"]
 
-Use a config.yaml file
-======================
+Use a YAML configuration file and allow for overrides
+=====================================================
 
-Configuration is data and should live inside a YAML file named
-``config.yaml``. You can access it in your Snakefile by including the
-line:
-
-.. code-block:: python
-
-   configfile: "config.yaml"
-
-and then using the ``config`` dictionary provided in scope afterwards.
-
-Allow for easy config overrides
-===============================
-
-By including the following snippet in your Snakefile, you can allow for
-optional configuration overrides using the ``--configfile`` option to
-``snakemake``.
+Configuration is data and should live inside YAML files. By including the
+following snippet in your Snakefile, you can provide default values and allow
+for additional entries or overrides via the ``--configfile`` or ``--config``
+options to ``snakemake``.
 
 .. code-block:: python
 
-   if not config:
-       configfile: "config.yaml"
+   configfile: "defaults.yaml"
+
+Configuration values are available as a ``config`` dictionary provided in scope
+afterwards.
 
 Use Snakemake ``params:`` block to map into ``config`` dictionary
 =================================================================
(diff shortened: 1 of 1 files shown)


src/reference/snakemake-style-guide.rst
```diff
@@ -46,3 +46,143 @@ Avoid the ``message`` rule attribute
...
+
+::
+
+   from pathlib import Path
+
+   if Path("config_local.yaml").is_file():
+       configfile: "config_local.yaml"
+
+If you do this, you should also add a ``/config_local.yaml`` line to
+your repository's top-level ``.gitignore`` file.
```
tsibley, line 78: This should also probably be dropped.

src/reference/snakemake-style-guide.rst
```diff
@@ -46,3 +46,143 @@ Avoid the ``message`` rule attribute
...
+Always use relative paths
+=========================
+
+Relative paths (paths that don't start with ``/``) mean that anyone can
+run the build without running into portability issues caused by paths
+specific to your computer.
+
+Our convention for builds is that Nextstrain components are available
+locally if needed as sibling directories, for example ``../fauna/`` and
+``../auspice/``.
```
tsibley, line 172: This ¶ should probably be dropped given that most builds don't/won't use Fauna and Auspice is available as the `auspice` command now.

tsibley (changes requested):
Thanks for doing this!

src/reference/snakemake-style-guide.rst
joverlee521: Maybe replace with the current standard that we use in [monkeypox](https://github.com/nextstrain/monkeypox/blob/5c461dc7e90cd70c1f16b193f82fd1666d4c95e2/Snakefile#L20-L22):

```python
if not config:
    configfile: "config/config_hmpxv1.yaml"
```

src/reference/snakemake-style-guide.rst
```diff
@@ -46,3 +46,127 @@ Avoid the ``message`` rule attribute
...
+           --output-sequences {output.sequences:q} \
+           --output-metadata {output.metadata:q}
+       """
+
+Always use relative paths
+=========================
+
+Relative paths (paths that don't start with ``/``) mean that anyone can
+run the build without running into portability issues caused by paths
+specific to your computer.
```
joverlee521, line 168: Ah, maybe link to [Snakemake docs on relative paths](https://snakemake.readthedocs.io/en/stable/project_info/faq.html#how-does-snakemake-interpret-relative-paths) since they are interpreted differently for different directives in Snakemake.

victorlin: Done: 9414b01774c8289257dfd8477b27ac330a71b95c

src/reference/snakemake-style-guide.rst
victorlin: Done: 295cbb7abc237939821bc8c9bb291468ed2a197a

src/reference/snakemake-style-guide.rst
victorlin: Done: b0c50da30f3e55e099199cc13e806bf12a5e0015

src/reference/snakemake-style-guide.rst
```diff
@@ -61,3 +61,122 @@ global variable. 3 ways are supported, but only 2 should be used:
...
+===============================
+
+By including the following snippet in your Snakefile, you can allow for
+optional configuration overrides using the ``--configfile`` option to
+``snakemake``.
+
+.. code-block:: python
+
+   if not config:
+       configfile: "config.yaml"
```
tsibley, line 88: These recommendations seems at odds with each other. Which should it be?

victorlin: This is a good point. I'm not sure since [ncov](https://github.com/nextstrain/ncov/blob/d17782f5ea422052c6ec6d2cc03ec8346ebdee0c/Snakefile#L38) uses the former and [monkeypox](https://github.com/nextstrain/monkeypox/blob/12ccb428aabccd9271fe8c5e61533edd656474e4/Snakefile#L14-L16) uses the latter.

If I had to choose, I would go with the former because it allows for partial overrides through Snakemake CLI parameters. The latter requires `config` to be all-or-nothing.

I will try creating a PR in the monkeypox repo using the former approach to see if it makes sense to recommend that here.

huddlej: I agree that we should use the first example in the snippet above, since that allows the entire config file to be overridden from the command line with `--configfile` and it allows specific config variables from the default config to be overridden from the command line.

Checking for the presence of an existing global config like the second block in the snippet is something we might do in a workflow-specific manner especially for workflows that allow users to define custom rules, but I wouldn't recommend that as the standard default behavior.

victorlin: Updated in 7fb0ecbfdd312a3f10ca531510092fa1066e7f77. Applied to the monkeypox repo: https://github.com/nextstrain/monkeypox/pull/173

tsibley (approved):
I pushed three additional commits instead of doing another round of review/changes. Take a gander and see if you agree?

victorlin:
@tsibley thanks for doing that, changes look good! Merging.

PR #167, opened 2023-08-10
install: Recommend WSL more explicitly among Windows options

victorlin:
_Install page: [current](https://docs.nextstrain.org/en/latest/install.html), [preview](https://nextstrain--167.org.readthedocs.build/en/167/install.html)_

Currently, when choosing between the two Windows options, we recommend the WSL interface to Nextstrain as noted in the FAQ page. However, I don't think the installation page reflects this recommendation well enough. To an inexperienced user, the PowerShell interface (named "Windows") might seem like a better option to choose over "WSL on Windows", which conflicts with the recommendation in FAQ.

This PR renames those options to better reflect the recommendation directly on the install page.

### Related issue(s)

N/A

### Testing

- [x] Checks pass

### DIFF
--- a/src/install.rst
+++ b/src/install.rst
@@ -52,6 +52,7 @@ Install Nextstrain CLI
          Most of the time, Rosetta 2 will already be enabled.
          If not, the installer will ask you to first enable Rosetta 2 and then retry the installation.
 
+
    .. group-tab:: Windows (PowerShell)
 
       In a PowerShell terminal, run:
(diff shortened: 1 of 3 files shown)


j23414 (approved):
Like the reordered tabs and wording changes, and [LGTM](https://nextstrain--167.org.readthedocs.build/en/167/install.html)!

victorlin:
I force-pushed a few tweaks while taking another pass. Merging now.

PR #170, opened 2023-09-14
Snakemake style guide: Add another way of accessing config values

victorlin:

Checking presence of a key in the config should be done with the membership operator. This must be done otherwise false-y values will stay false when membership is checked with `config.get`.

## Checklist

- [x] Checks pass


victorlin:
Merging this small change. Open to reverting if anyone feels it's unnecessary.

PR #173, opened 2023-11-14
Add log recommendations to Snakemake style guide

huddlej:

Adds two recommendations about logging to the Snakemake style guide based on recent difficulty troubleshooting workflow issues with users.

## Checklist

- [x] Checks pass

### DIFF
--- a/src/reference/snakemake-style-guide.rst
+++ b/src/reference/snakemake-style-guide.rst
@@ -193,3 +193,36 @@ Example:
            --output-sequences {output.sequences:q} \
            --output-metadata {output.metadata:q}
        """
+
+Log standard out and error output to log files and the terminal
+===============================================================
+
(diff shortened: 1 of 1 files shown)


src/reference/snakemake-style-guide.rst
```diff
@@ -193,3 +193,36 @@ Example:
...
+       log:
+           "logs/filter.txt"
+       shell:
+           """
+           augur filter \
+               --metadata {input.metadata} \
+               --output-metadata {output.metadata} 2>&1 | tee {log}
+           """
+
+Before using ``tee``, ensure that your workflow uses bash's ``pipefail`` option, so successful ``tee`` execution does not mask errors from earlier commands in the pipe.
```
jameshadfield, line 221: We should document how to do this (I don't see it in this style guide)

huddlej: Good call. Since the pipefail option is enabled by default in Snakemake, we shouldn't have to do anything but check for parts of the workflow that override the defaults. I added examples of how to check that the defaults are overridden and also a link to the definition of the "bash strict mode".

PR #185, opened 2024-01-31
Add instructions for singularity runtime

victorlin:

Descriptive snippet copied from Nextstrain CLI docs.

<https://github.com/nextstrain/cli/blob/6da49c541/nextstrain/cli/runner/singularity.py#L4-L9>

## Related issue(s)

- Suggested in https://github.com/nextstrain/cli/pull/353#issuecomment-1919675153

## Checklist

- [x] Checks pass

### DIFF
--- a/src/install.rst
+++ b/src/install.rst
@@ -184,6 +184,29 @@ Set up a Nextstrain runtime
             .. include:: snippets/nextstrain-setup-conda.rst
 
 
+   .. group-tab:: Singularity (Apptainer)
+
+      Singularity is a container system freely-available for Linux platforms. It is commonly available on institutional HPC systems as an alternative to Docker, which is often not supported on such systems.
+
+      The Singularity project forked into two separate projects in late 2021: `SingularityCE <https://sylabs.io/singularity/>`__ and `Apptainer <https://apptainer.org>`__.  Either fork should work with Nextstrain CLI, as both projects still provide very similar interfaces and functionality via the ``singularity`` command.
+
+      .. note::
+
+         These instructions are for institutional HPC systems that have Singularity installed and configured. You may need to ``module load`` SingularityCE or Apptainer. If you don't have Singularity available, consider using one of the other runtimes or continue by installing `SingularityCE <https://docs.sylabs.io/guides/3.0/user-guide/installation.html>`__/`Apptainer <https://apptainer.org/docs/admin/main/installation.html>`__.
+
+      1. Check that Singularity is available in your environment.
+
+         .. code-block:: bash
+
+            singularity --version
+
+      2. Set up the runtime:
+
+         .. code-block:: none
+
+            nextstrain setup --set-default singularity
+
+
    .. group-tab:: Ambient (advanced)
 
       .. We use the phrase "custom Conda environment" to refer to a Conda environment managed by the user for use with the ambient runtime.
@@ -265,6 +288,33 @@ Try running Augur and Auspice
 
 
    .. group-tab:: Conda
+
+      1. Enter an interactive Nextstrain shell in the current directory (``.``).
+
+         .. code-block:: bash
+
+            nextstrain shell .
+
+      2. Run Augur.
+
+         .. code-block:: bash
+
+            augur --help
+
+      3. Run Auspice.
+
+         .. code-block:: bash
+
+            auspice --help
+
+      4. Exit the Nextstrain shell.
+
+         .. code-block:: bash
+
+            exit
+
+
+   .. group-tab:: Singularity (Apptainer)
 
       1. Enter an interactive Nextstrain shell in the current directory (``.``).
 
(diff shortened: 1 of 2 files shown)


src/install.rst
```diff
@@ -184,6 +184,23 @@ Set up a Nextstrain runtime
...
 
+   .. group-tab:: Singularity
+
+      Singularity is a container system freely-available for Linux platforms. It is commonly available on institutional HPC systems as an alternative to Docker, which is often not supported on such systems.
+
+      1. Check that Singularity is available in your environment.
+
+         .. code-block:: bash
+
+            singularity --version
```
tsibley, line 195: The Docker section in this doc describes how to install/configure Docker itself on different OS/platforms. We're assuming that's already done for Singularity, I guess? and that it doesn't differ per OS/platform?

I think I'd be more inclined to replicate the second-level OS/platforms tabs, and guide folks to using Singularity on Windows and macOS where using it requires a VM (WSL2 or otherwise). This is what we do for the Conda runtime, after all.

src/install.rst
```diff
@@ -184,6 +184,23 @@ Set up a Nextstrain runtime
             .. include:: snippets/nextstrain-setup-conda.rst
 
 
+   .. group-tab:: Singularity
+
+      Singularity is a container system freely-available for Linux platforms. It is commonly available on institutional HPC systems as an alternative to Docker, which is often not supported on such systems.
```
tsibley, line 189: As @j23414 [suggested elsewhere](https://github.com/nextstrain/cli/pull/353#issuecomment-1919803164), it's probably worth noting here that Singularity refers to both forks: SingularityCE and Apptainer (at least as long as a `singularity` executable is available). There's language to crib from at https://docs.nextstrain.org/projects/cli/en/stable/runtimes/singularity/?highlight=Apptainer.

src/install.rst
tsibley: Or, if we want to assume Singularity is only being used in places where an administrator has installed and configured it, that we note as much and suggest that it might need enablement by the HPC's modules system (e.g. `modules load singularity` or something).

victorlin: It's easier for us to assume Singularity is used as a Docker replacement for HPC systems. That seems to be where most of the users will come from, at least for now. I've [updated](https://github.com/nextstrain/docs.nextstrain.org/compare/cb4e30fa1d26fbd8761622785db5dc20a91fdc64..0961b9eb3564f7f776c1791c8a27e3e5e5f29567) wording to make that explicit.

src/install.rst
victorlin: [Updated](https://github.com/nextstrain/docs.nextstrain.org/compare/cb4e30fa1d26fbd8761622785db5dc20a91fdc64..0961b9eb3564f7f776c1791c8a27e3e5e5f29567) to mention SingularityCE/Apptainer.

src/install.rst
tsibley: This is great, thanks.

src/install.rst
```diff
@@ -184,6 +184,29 @@ Set up a Nextstrain runtime
             .. include:: snippets/nextstrain-setup-conda.rst
 
 
+   .. group-tab:: Singularity (Apptainer)
```
tsibley, line 187: I was reviewing the rendered docs and realized that we're missing this tab in the "[Try running Augur and Auspice](https://nextstrain--185.org.readthedocs.build/en/185/install.html#try-running-augur-and-auspice)" section of the install page. Singularity is also missing from the "[Managing an installation](https://nextstrain--185.org.readthedocs.build/en/185/guides/manage-installation.html)" page. Let's add it there too?

src/install.rst
```diff
@@ -184,6 +184,29 @@ Set up a Nextstrain runtime
...
+
+         .. code-block:: bash
+
+            singularity --version
+
+      2. Set up the runtime:
+
+         .. code-block:: none
+
+            nextstrain setup --set-default singularity
```
victorlin, line 207: Tested on Fred Hutch's `rhino` and it didn't work. I created a separate issue: https://github.com/nextstrain/cli/issues/360

src/install.rst
victorlin: Not sure what to add under "Troubleshoot a broken installation". I tried setting it up on `rhino` to get a sense of what to put here, but that was [unsuccessful](https://github.com/nextstrain/docs.nextstrain.org/pull/185/files#r1476544501)). Any ideas?

tsibley: > Not sure what to add under "Troubleshoot a broken installation"

I think something relatively generic like what the Conda tab uses would be ok.

victorlin: Added in [force-push](https://github.com/nextstrain/docs.nextstrain.org/compare/0961b9eb3564f7f776c1791c8a27e3e5e5f29567..8b29c2f846358612ed461d5cf2456ec9de4e56ae)

tsibley (approved):

Issue #188, opened 2024-02-07
Documentation for creating a pathogen repo


joverlee521:
There are existing resources that we can point to as starting point:
- [Alli's Epi Handbook](https://alliblk.github.io/genepi-book/) 
- [Nextstrain VEME tutorial](https://github.com/blab/veme-2022) 
- [pathogen-repo-guide](https://github.com/nextstrain/pathogen-repo-guide) 

However, most likely users will require additional guidance on general decisions as they set up their pathogen repo. 
I imagine this to be a high level decision tree that can link out to other detailed docs. 

Jotting down general questions and thoughts here that I think would fit into the decision tree:

- Do I need to create my own pathogen repo, use Nextclade, or fork an existing Nextstrain repo?
    - Is there an example for _xyz_ pathogen? (https://docs.nextstrain.org/en/latest/guides/bioinformatics/examples.html#nextstrain-pathogens) 
- Do I need to create an ingest workflow? 
    - How do I create an ingest workflow? (https://github.com/nextstrain/docs.nextstrain.org/issues/179)
- Do I need to create a Nextclade dataset?
    - How do I create a Nextclade dataset? ([Nextclade dataset creation guide](https://github.com/nextstrain/nextclade_data/blob/8071aa0fe860e085b0593fc6d17e6804ac5c2b5a/docs/dataset-creation-guide.md))
- How do I create a phylogenetic workflow?
    - How do I choose a reference? (https://docs.nextstrain.org/en/latest/guides/bioinformatics/translate_ref.html)
    - How do I treat gaps? (https://docs.nextstrain.org/en/latest/guides/bioinformatics/missing-sequence-data.html)
    - How do I choose an aligner?
    - How do I decide which params to use for Augur commands?
    - Should I use BEAST or TreeTime?
    - How do I choose a subsampling scheme? (https://alliblk.github.io/genepi-book/sample-selection.html)
    - How do I implement my subsampling scheme? (e.g. https://github.com/nextstrain/docs.nextstrain.org/issues/187)

tsibley:
FWIW, I'd think to co-locate the docs for this topic within the pathogen-repo-guide repo itself. It seems like a nice central place to put them and they can stay in lock step with changes to the layout/organization in the guide.

joverlee521:
[...]

PR #191, opened 2024-03-02
Update glossary and graphviz for repo/workflows

### LINKED ISSUE #190: Update glossary terms `workflow` and `workflow repository`,
The current definition for `workflow` and `workflow repository` is outdated considering a pathogen repository can have up to 3 different directories with individual workflows (ingest, phylogenetic, nextclade). 

https://github.com/nextstrain/docs.nextstrain.org/blob/2ca0b7266ccdb86bf27dd91f774071fdc133ea0e/src/reference/glossary.rst?plain=1#L15-L28

https://github.com/nextstrain/docs.nextstrain.org/blob/2ca0b7266ccdb86bf27dd91f774071fdc133ea0e/src/reference/glossary.rst?plain=1#L33-L36

joverlee521:

With the push for pathogen repos to adhere to the pathogen-repo-guide, the glossary and graphviz for repositories and workflows need to be updated.

This will also make it easier to standardize terminology in the upcoming ingest tutorials.

## Related issue(s)

Resolves #190 

## Checklist

- [ ] Checks pass

### DIFF
--- a/src/reference/glossary.rst
+++ b/src/reference/glossary.rst
@@ -63,8 +63,7 @@ Glossary
    build
       also *Nextstrain build*
 
-      *(noun)* A sequence of commands, parameters and input files which work together to reproducibly generate outputs.
-      Phylogenetic builds execute bioinformatic analyses and generate a :term:`dataset` for visualization with :term:`Auspice`.
+      *(noun)* A sequence of commands, parameters and input files which work together to reproducibly generate a :term:`dataset`.
 
    build (verb)
 
@@ -118,7 +117,7 @@ Glossary
 
    narrative
 
-      A method of data-driven storytelling with interactive views of :term:`datasets <dataset>` displayed alongside multiple pages (or slides) of text and images.
+      A method of data-driven storytelling with interactive views of :term:`phylogenetic datasets<phylogenetic dataset>` displayed alongside multiple pages (or slides) of text and images.
       Saved as a Markdown file with extended syntax to support additional displays.
 
       Viewable on nextstrain.org or with :term:`Auspice` via the :doc:`cli:commands/view` or :doc:`auspice view <auspice:introduction/how-to-run>` commands.
@@ -126,7 +125,7 @@ Glossary
       See also :doc:`/guides/communicate/narratives-intro` and :doc:`/tutorials/narratives-how-to-write`.
 
    JSONs
-      Special ``.json`` files produced by :term:`Augur` and visualized by :term:`Auspice`. These files make up a :term:`dataset`.
+      Special ``.json`` files produced by :term:`Augur` and visualized by :term:`Auspice`. These files make up a :term:`phylogenetic dataset`.
       See :doc:`data formats<data-formats>`.
 
    Nextstrain CLI
--- a/src/reference/glossary.rst
+++ b/src/reference/glossary.rst
@@ -19,9 +19,14 @@ Glossary
    workflow
       also *Nextstrain workflow*
 
-      A reproducible process comprised of one or more :term:`builds<build>` producing outputs.
-      The outputs produced are often :term:`datasets<dataset>`, which can be visualized by :term:`Auspice`.
+      A reproducible process comprised of one or more :term:`builds<build>` producing :term:`datasets<dataset>`.
       Implementation varies per workflow, but generally they are run by workflow managers such as Snakemake.
+
+      A Nextstrain :term:`pathogen repository` typically consists of these different workflows
+
+      1. :term:`phylogenetic workflow`
+      2. :term:`ingest workflow`
+      3. :term:`Nextclade workflow`
 
       Our :term:`core workflows<core workflow>` can be divided into two types:
 
@@ -31,6 +36,25 @@ Glossary
       .. note::
 
          The individual builds in a multi-build workflow are also "workflows" in the definition of workflow managers like Snakemake.
+
+   phylogenetic workflow
+
+      A :term:`workflow` consisting of :term:`build(s)<build>` that execute bioinformatic analyses with :term:`Augur` to generate
+      :term:`phylogenetic dataset(s)<phylogenetic dataset>` for visualization with :term:`Auspice`.
+
+      The phylogenetic workflow is often considered the primary workflow in a pathogen repository
+      (e.g. "the Zika workflow" typically means "the phylogenetic workflow in the Zika pathogen repository").
+
+   ingest workflow
+
+      A :term:`workflow` consisting of :term:`build(s)<build>` that curate public metadata and sequences to generate
+      :term:`ingest dataset(s)<ingest dataset>` that are typically used as input files for
+      :term:`phylogenetic workflows<phylogenetic workflow>` and :term:`Nextclade workflows<Nextclade workflow>`.
+
+   Nextclade workflow
+
+      A :term:`workflow` consisting of :term:`build(s)<build>` that generate :doc:`reference tree(s)<nextclade:user/input-files/04-reference-tree>` to be packaged with other
+      dataset files to create :term:`Nextclade dataset(s)<Nextclade dataset>`.
 
    core workflow
 
--- a/src/reference/glossary.rst
+++ b/src/reference/glossary.rst
@@ -61,7 +61,7 @@ Glossary
       A :term:`workflow` maintained by the Nextstrain team.
 
    build
-      also *Nextstrain build*
+      also *Nextstrain build*, *phylogenetic build*, *ingest build*, *Nextclade build*
 
       *(noun)* A sequence of commands, parameters and input files which work together to reproducibly generate a :term:`dataset`.
 
--- a/src/reference/glossary.rst
+++ b/src/reference/glossary.rst
@@ -75,20 +75,46 @@ Glossary
       A modular instruction of a :term:`build` which can be run standalone (e.g. ``augur filter``), often with clear input and output files.
 
    dataset
+
+      A collection of output files produced by a :term:`build`.
+      A Nextstrain :term:`pathogen repository` typically produces multiple types of datasets
+
+      1. :term:`phylogenetic dataset`
+      2. :term:`ingest dataset`
+      3. :term:`Nextclade dataset`
+
+   phylogenetic dataset
       also *Auspice JSONs*
 
-      A collection of :term:`JSONs` produced by a :term:`build`. It is also the shared file prefix of the JSONs. For example ``flu/seasonal/h3n2/ha/2y`` identifies a dataset which corresponds to the files
-      :
+      A :term:`dataset` consisting of :term:`JSONs` produced by a :term:`build` of a :term:`phylogenetic workflow`.
+      It is also the shared file prefix of the JSONs.
+      For example ``flu/seasonal/h3n2/ha/2y`` identifies a dataset which corresponds to the files:
 
       - ``flu_seasonal_h3n2_ha_2y_meta.json``
       - ``flu_seasonal_h3n2_ha_2y_tree.json``
       - ``flu_seasonal_h3n2_ha_2y_tip-frequencies.json``
 
-      Some :term:`phylogenetic workflows<phylogenetic workflow>` produce a single, synonymous dataset, like Zika. Others, like seasonal flu, produce many datasets.
+      Some phylogenetic workflows produce a single, synonymous dataset, like Zika. Others, like seasonal flu, produce m
+
+      The phylogenetic dataset is often considered the primary dataset in a pathogen repository
+      (e.g. "the Zika dataset" typically means "the phylogenetic dataset from the Zika pathogen repository").
+
+   ingest dataset
+
+      A :term:`dataset` consisting of curated files produced by a :term:`build` of an :term:`ingest workflow`.
+      Typically consists of the files:
+
+      * metadata.tsv
+      * sequences.fasta
+
+      If the ingest workflow includes Nextclade :term:`build steps<build step>`, then the dataset will typically include
+      :doc:`Nextclade output files<nextclade:user/output-files/index>` as well.
 
    Nextclade dataset
 
-      A collection of input data files required for :doc:`Nextclade<nextclade:index>` to run an analysis. :doc:`Documentation<nextclade:user/datasets>`
+      A :term:`dataset` consisting of files required for a :doc:`Nextclade<nextclade:index>` analysis, usually produced
+      by a :term:`build` of a :term:`Nextclade workflow`.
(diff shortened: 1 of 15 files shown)


joverlee521:
Main changes are
- Parts of a whole page: [current](https://docs.nextstrain.org/en/latest/learn/parts.html) vs [preview](https://nextstrain--191.org.readthedocs.build/en/191/learn/parts.html)
- Glossary page: [current](https://docs.nextstrain.org/en/latest/reference/glossary.html) vs [preview](https://nextstrain--191.org.readthedocs.build/en/191/reference/glossary.html)

Leaving this as a draft for discussion.

src/reference/glossary.rst
```diff
@@ -30,15 +36,11 @@ Glossary
...
-      also *pathogen workflow repository*
-
-      A version-controlled folder containing all files necessary to run a :term:`workflow`.
-
    build
       also *Nextstrain build*
 
-      *(noun)* A sequence of commands, parameters and input files which work together to reproducibly execute bioinformatic analyses and generate a :term:`dataset` for visualization with :term:`Auspice`.
+      *(noun)* A sequence of commands, parameters and input files which work together to reproducibly generate outputs.
+      Phylogenetic builds execute bioinformatic analyses and generate a :term:`dataset` for visualization with :term:`Auspice`.
```
victorlin, line 44: Do you see ingest/nextclade as other types of builds, analagous to the types of workflows?

src/reference/glossary.rst
```diff
@@ -12,10 +12,16 @@ Glossary
 
       A web application used for phylogenetic visualization and analysis. :doc:`Documentation<auspice:index>`
 
+   pathogen repository
+
+      A version-controlled folder containing all files necessary to run a pathogen's :term:`workflows<workflow>`.
+
    workflow
-      also *pathogen workflow*, *pathogen analysis*, *Nextstrain workflow*
```
victorlin, line 16: The tutorial titles "Running a pathogen workflow" / "Creating a pathogen workflow" no longer have a clear definition of "pathogen workflow". The definition changes are getting rid of the previously assumed mapping of 1 repo : 1 workflow which is also reflected in changes to the zika repo. (Suggestion 1) This would be a good opportunity to replace zika-tutorial with zika as mentioned elsewhere.

(Suggestion 2) Rework the tutorials to account for the multi-workflow repo:

> Title: Using a pathogen repository
> 
> 1. Clone zika repo
> 2. Explain the different workflows
> 3. Assume ingest/nextclade are not necessary and show how to run the phylo workflow (existing content)

> Title: Creating a pathogen repository
> 
> 1. Show how to run a build by calling Augur directly (existing content)
> 2. Explain the scoping of `phylogenetic/`

Maybe it'd be useful to have an additional tutorial for users who want to run without the rigid pathogen repo layout, e.g. "Running a phylogenetic analysis with Augur commands and visualizing with Auspice", where contents would be what's in the current "Creating a pathogen workflow" tutorial.

victorlin (commented):
Thanks for starting this PR, these changes are much needed! The suggestions below are based on the idea of going full send in the direction of multi-workflow pathogen repos, instead of keeping existing wording that assumes each pathogen repo = 1 phylo workflow. This means drawing clear boundaries around the types of workflows to align with pathogen-repo-guide, and at least acknowledging the presence of ingest/nextclade workflows if not walking through them in detail like the phylo workflow.

src/reference/glossary.rst
joverlee521: Yes? To me, the terms `build` and `workflow` really start to blend together here...If we define the 3 different types of workflows [as you suggested](https://github.com/nextstrain/docs.nextstrain.org/pull/191#discussion_r1511690095), then I think we can remove the specific definitions for types of builds.

src/reference/glossary.rst
joverlee521: Thanks! I think both suggestions fit into the long term goal to have a thorough [docs for creating a pathogen repo](https://github.com/nextstrain/docs.nextstrain.org/issues/188). 

I think I'd just rename the two tutorial titles to "Running a phylogenetic workflow" and "Creating a phylogenetic workflow" but incorporate the other suggestions in future work. I think the addition of the ingest tutorials will require some restructuring of the existing tutorials anyways.

src/reference/glossary.rst
joverlee521: Ah, I see! Yes, ingest/nextclade both can have this distinction as well. 

For nextclade, the mpox nextclade workflow has [3 different builds](https://github.com/nextstrain/mpox/blob/e439235ff1c1d66e7285b774e9536e2896d9cd2f/nextclade/Snakefile#L6) that produces a different [mpox Nextclade dataset](https://github.com/nextstrain/nextclade_data/tree/master/data/nextstrain/mpox).

For ingest, this might not be as clear. For example, the [dengue ingest workflow](https://github.com/nextstrain/dengue/tree/main/ingest#usage) produces `all`, `denv_1`, `denv_2`, `denv_3`, `denv_4` data files, where denv 1-4 are different serotypes that are subsets of the full data. I think each serotype can be considered a different build?

victorlin: Got it! In that case, it might also make sense to generalize the term **dataset** to be the output of the build process:

1. Phylogenetic dataset = the output of an Augur build (to be used by Auspice)
2. Nextclade dataset = the output of a Nextclade build
3. Ingest dataset = the output of an ingest build (e.g. for a specific dengue serotype)

joverlee521: I think generalizing the term `dataset` is a good direction, but I'm also hesitant because it's so closely tied to Auspice JSONs in all of our docs and code. I guess we can add a special note to the term `dataset`:

```
Unless otherwise specified, a "dataset" usually refers to a set of Auspice JSONs
```

victorlin: 👍 special note would be good. Similar to this note from [another suggestion](https://github.com/nextstrain/docs.nextstrain.org/pull/191#discussion_r1511690095):

> Often considered the primary workflow in a pathogen repository (e.g. "the Zika workflow" typically means "the phylogenetic workflow in the Zika pathogen repository").

src/reference/glossary.rst
```diff
@@ -12,10 +12,16 @@ Glossary
...
+      also *Nextstrain workflow*
 
-      A reproducible process comprised of one or more :term:`builds<build>` producing :term:`datasets<dataset>`, which can be visualized by :term:`Auspice`. Implementation varies per workflow, but generally they are run by workflow managers such as Snakemake.
+      A reproducible process comprised of one or more :term:`builds<build>` producing outputs.
+      The outputs produced are often :term:`datasets<dataset>`, which can be visualized by :term:`Auspice`.
+      Implementation varies per workflow, but generally they are run by workflow managers such as Snakemake.
```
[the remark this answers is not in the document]
joverlee521: Updated in 6ad02eed77f2416acfd85d57e4c5dbf749b73e75

src/reference/glossary.rst
joverlee521: Updated in 7a7b50c1b422be2f06b031844bd31e388c8c78fe

src/reference/glossary.rst
joverlee521: Updated in cb76eebaa60a0a59d8549be264c9260e12254a25

joverlee521:
Major updates after initial feedback from @victorlin: 

- Generalization of the terms [workflow](https://nextstrain--191.org.readthedocs.build/en/191/reference/glossary.html#term-workflow) and [dataset](https://nextstrain--191.org.readthedocs.build/en/191/reference/glossary.html#term-dataset)
- Addition of individual workflow terms: [phylogenetic workflow](https://nextstrain--191.org.readthedocs.build/en/191/reference/glossary.html#term-phylogenetic-workflow), [ingest workflow](https://nextstrain--191.org.readthedocs.build/en/191/reference/glossary.html#term-ingest-workflow), and [Nextclade workflow](https://nextstrain--191.org.readthedocs.build/en/191/reference/glossary.html#term-Nextclade-workflow)
- Addition of individual dataset terms: [phylogenetic dataset](https://nextstrain--191.org.readthedocs.build/en/191/reference/glossary.html#term-phylogenetic-dataset), [ingest dataset](https://nextstrain--191.org.readthedocs.build/en/191/reference/glossary.html#term-ingest-dataset), and [Nextclade dataset](https://nextstrain--191.org.readthedocs.build/en/191/reference/glossary.html#term-Nextclade-dataset). 

I'm planning to add the ingest tutorials in a separate branch based on these changes so they can link back to these new terms.

j23414 (approved):
I can get behind the use of the terms. As far as I understand, I'd just need to practice the new phrasing:
* “phylogenetic analysis” -> “phylogenetic workflow” (to focus on the code)
* “phylogenetic build” -> “phylogenetic dataset” (to focus on the final files)

src/reference/glossary.rst
```diff
@@ -12,10 +12,21 @@ Glossary
 
       A web application used for phylogenetic visualization and analysis. :doc:`Documentation<auspice:index>`
 
+   pathogen repository
+
+      A version-controlled folder containing all files necessary to run a pathogen's :term:`workflows<workflow>`.
+
    workflow
-      also *pathogen workflow*, *pathogen analysis*, *Nextstrain workflow*
+      also *Nextstrain workflow*
```
victorlin, line 20: Commenting on this unmodified part:

https://github.com/nextstrain/docs.nextstrain.org/blob/8c7779adefeaab4851726468b98223349a5c205a/src/reference/glossary.rst#L59-L61

Suggestion: Shift from *core workflow* to *core repository* to align with the shift from *pathogen workflow* to *pathogen repository*. We maintain repos in their entirety, not just parts of the repo.

src/reference/glossary.rst
victorlin: The addition of individual terms for **workflow** and **dataset** seems incomplete without doing the same for **build**.

Suggestion: add individual terms for *phylogenetic/ingest/Nextclade build* as described in https://github.com/nextstrain/docs.nextstrain.org/pull/191#discussion_r1511939744

src/reference/glossary.rst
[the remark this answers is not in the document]
victorlin: Maybe a flat glossary page isn't the best way to describe these ambiguous terms. Here's an alternative using tabs: 7293090bd1bee4796b87106e51b32370f9752fea

victorlin (changes requested):

src/reference/glossary.rst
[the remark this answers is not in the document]
joverlee521: I actually really like this bit to help distinguish workflows and builds. It also helps with explaining why some workflows are more complicated than others.

src/reference/glossary.rst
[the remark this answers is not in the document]
joverlee521: Using tabs seems like a great way to organize these terms! I'm a bit wary that there's no way to link to a specific tab. It seemed handy to reference these terms in the [upcoming ingest tutorial](https://github.com/nextstrain/docs.nextstrain.org/blob/79a0176a4a76e120ebefb42afa90fb073c7d46fc/src/tutorials/using-a-pathogen-repo/running-an-ingest-workflow.rst?plain=1#L5-L7).

src/reference/glossary.rst
joverlee521: Hmm, I didn't add individual terms for build because I had thought the [current definition](https://nextstrain--191.org.readthedocs.build/en/191/reference/glossary.html#term-build) applies across all three.

I'm having trouble differentiating the phylogenetic/ingest/Nextclade build terms. 
This seems a bit overkill to me, but you probably have a better suggestion? 

```rst

phylogenetic build 

    (noun) A sequence of commands, parameters and input files which work together to 
    reproducibly generate a phylogenetic dataset. Phylogenetic builds are often have distinct 
    input files and subsampling schemes for slightly different bioinformatic analyses.

ingest build
    
    (noun) A sequence of commands, parameters and input files which work together to 
    reproducibly generate an ingest dataset. Ingest builds often have distinct upstream data 
    sources and groupings by pathogen subtypes or genes.

nextclade build

    (noun) A sequence of commands, parameters and input files which work together to 
    reproducibly generate a Nextclade dataset. Nextclade builds often have distinct 
    reference sequences and subsampling schemes to be used for different scopes of Nextclade analyses.
```

src/reference/glossary.rst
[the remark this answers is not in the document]
victorlin: Ok, let's keep it then!

victorlin: Nice, I hadn't considered that framing. It aligns with the existing usage of the term.

src/reference/glossary.rst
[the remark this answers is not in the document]
victorlin: Yeah, it's a bummer they can't be linked (`:ref:` doesn't work with anchors within tabs).

[sphinx-inline-tabs](https://github.com/pradyunsg/sphinx-inline-tabs/pull/30) supports linking to specific tabs but it's a bit ugly (`…/glossary.html?tabs=Nextclade#term-workflow`).

Probably best to keep as-is without tabs.

src/reference/glossary.rst
victorlin: Oh you're right, it does seem overkill. Maybe better to stick with a single entry with more specific aliases:

> **build**
>
>  also *Nextstrain build*, *phylogenetic build*, *ingest build*, *Nextclade build*

src/reference/glossary.rst
```diff
@@ -12,10 +12,21 @@ Glossary
...
+
```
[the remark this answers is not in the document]
joverlee521: Hmm yeah, often a "Nextstrain workflow" is just referring to a phylogenetic workflow. I'll move this to be an alias for phylogenetic workflow.

src/reference/glossary.rst
joverlee521: Updated in https://github.com/nextstrain/docs.nextstrain.org/pull/191/commits/63b716bd1ac64d18215fc42ce85ae0b6ae8a23bc

src/reference/glossary.rst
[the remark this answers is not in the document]
joverlee521: Updated in https://github.com/nextstrain/docs.nextstrain.org/pull/191/commits/fa99a2351ee1a210d5188dd6b911cdd38c3c5f28

src/reference/glossary.rst
joverlee521: Updated in https://github.com/nextstrain/docs.nextstrain.org/pull/191/commits/1ce52a7b80e52f634c2c984736cbeaf32a36e082

tsibley:
This is great! Nice work.

PR #192, opened 2024-03-06
Add guide on filtering and subsampling

### LINKED ISSUE #187: Add docs for "tiered" subsampling
A common question we get asked by users is how to do "tiered" subsampling, i.e. they want to use different subsampling parameters for `augur filter` for different geolocations. For example, a local public health department may want to have different subsampling parameters for global, country, neighboring states, and local sequences. 

The docs should aim to explain both the concept of tiered subsampling and provide examples of accomplishing this with Snakemake. 

1. Show how this can be done manually with multiple invocations of `augur filter` that does subsampling + a final invocation of `augur filter` that uses `--exclude-all` and `--include` flags to get the final set of sequences.
2. Show how this can be done in Snakemake with hard-coded invocations of `augur filter` per geolocation.
3. Show how this can be done in Snakemake with the [config-based subsampling approach](https://github.com/NW-PaGe/WNV-nextstrain/pull/1)

victorlin:

Uses a snippet from the Augur docs.

## Related issue(s)

- Closes #187
- Follow-up to https://github.com/nextstrain/augur/pull/1425

## Checklist

- [x] [Preview](https://nextstrain--192.org.readthedocs.build/en/192/guides/bioinformatics/index.html) looks good
- [x] Merge https://github.com/nextstrain/augur/pull/1431
- [x] Drop 1ac8b85e32be34ce46e3ad0f29325ccde51c42c3
- [ ] Checks pass

### DIFF
--- a/src/conf.py
+++ b/src/conf.py
@@ -82,6 +82,8 @@ intersphinx_mapping = {
     'augur': ('https://docs.nextstrain.org/projects/augur/page/', None),
     'auspice': ('https://docs.nextstrain.org/projects/auspice/page/', None),
     'cli': ('https://docs.nextstrain.org/projects/cli/page/', None),
+    # For externally sourced snippets that reference this docs project
+    'docs.nextstrain.org': ('https://docs.nextstrain.org/page/', None),
     'nextclade': ('https://docs.nextstrain.org/projects/nextclade/page/', None),
     'ncov': ('https://docs.nextstrain.org/projects/ncov/page/', None)
 }
--- a/src/guides/bioinformatics/index.rst
+++ b/src/guides/bioinformatics/index.rst
@@ -10,12 +10,12 @@ How-to guides for performing bioinformatic anaylses with Nextstrain.
    :titlesonly:
    :caption: Table of contents
 
-   defining-clades
    filtering-and-subsampling
    augur_snakemake
    missing-sequence-data
    translate_ref
    import-beast
+   defining-clades
    colors
    lat_longs
    vcf_input
--- a/src/guides/bioinformatics/index.rst
+++ b/src/guides/bioinformatics/index.rst
@@ -11,6 +11,7 @@ How-to guides for performing bioinformatic anaylses with Nextstrain.
    :caption: Table of contents
 
    defining-clades
+   filtering-and-subsampling
    augur_snakemake
    missing-sequence-data
    translate_ref
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ src/guides/bioinformatics/translate_ref.md
 src/guides/bioinformatics/vcf_input.md
 src/guides/communicate/create-pdf.md
 src/guides/communicate/narratives-intro.md
+src/snippets/filtering-and-subsampling.rst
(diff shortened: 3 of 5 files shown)


src/guides/bioinformatics/index.rst
```diff
@@ -21,3 +21,4 @@ How-to guides for performing bioinformatic anaylses with Nextstrain.
    fasta_input
    seq_traits
    examples
+   filtering-and-subsampling
```
victorlin, line 24: Open to suggestions for where to place this in the existing order. I don't think it should be at the very end, but unsure where would be best.

joverlee521: I'd vote to place it before the [Augur and snakemake page](https://nextstrain--192.org.readthedocs.build/en/192/guides/bioinformatics/augur_snakemake.html). 

The last part of the filtering/subsampling page that alludes to connecting to `augur align`

> If you are building a standard Nextstrain workflow, the output files should be used as input to sequence alignment

flows nicely into the example on the "Augur and snakemake" page that connects `augur filter` and `augur align`.

src/conf.py
```diff
@@ -82,6 +82,7 @@
     'augur': ('https://docs.nextstrain.org/projects/augur/page/', None),
     'auspice': ('https://docs.nextstrain.org/projects/auspice/page/', None),
     'cli': ('https://docs.nextstrain.org/projects/cli/page/', None),
+    'docs.nextstrain.org': ('https://docs.nextstrain.org/page/', None),
```
joverlee521, line 85: Took me a second to realize this needed to be added because the [augur/filter-and-subsampling.rst](https://github.com/nextstrain/augur/blob/b0c5010ec16744dbf8e51225b023d5ed1bc18ff6/docs/usage/cli/snippets/filtering-and-subsampling.rst?plain=1#L134) refers to this with `:doc:`

src/guides/bioinformatics/index.rst
```diff
@@ -11,6 +11,7 @@ How-to guides for performing bioinformatic anaylses with Nextstrain.
    :caption: Table of contents
 
    defining-clades
```
joverlee521, line 13: _non-blocking_

I'd consider moving `defining-clades` down and putting `filter-and-subsampling` first since that seems to be the top question from our users.

joverlee521 (approved):
Looks good to me 👍 

Thanks for following up to surface the filtering/subsampling docs on the top level!

src/guides/bioinformatics/index.rst
victorlin: Done: c942ae6dcf95a464228eeb02153157e4d62e4615

src/conf.py
victorlin: Added clarity in 96b3b9598105dc39a6f118cb752df606747d7191

src/guides/bioinformatics/index.rst
victorlin: Done: 3316ce674572bf2a551dfbc3025e8cfb8b41fa15

src/conf.py
tsibley: This will likely get weird at some point, as self-references like this means the references always use the published version, not the to-be-published version being built.

victorlin: Alternative proposed in https://github.com/nextstrain/augur/pull/1431#issuecomment-2067315935

Issue #179, opened 2024-01-18
Add tutorials for ingest pipelines


joverlee521:
Initially proposed by @trvrb in our meeting with WA DOH on 2024-01-17. 
The tentative due date for this objective is 2024-02-29.

We should have tutorials for running and creating ingest pipelines from NCBI data in parallel with the [Running a pathogen workflow](https://docs.nextstrain.org/en/latest/tutorials/running-a-workflow.html) and [Creating a pathogen workflow](https://docs.nextstrain.org/en/latest/tutorials/creating-a-workflow.html) tutorials. 

The ultimate goal would be for an external user to be able to follow the tutorials to set up their own ingest pipeline for any viral pathogen using NCBI data.

joverlee521:
As @kimandrews works through setting up ingest for measles, we can work on putting together a step by step tutorial for creating the ingest workflow.

joverlee521:
Jotting down some general thoughts to keep track of from conversations with @kimandrews 

- who is our intended audience? What is the expected level of experience with Nextstrain? 
- will need guide for installing `git subrepo` for managing the `ingest/vendored` scripts 
- document commands to get the first fetch from NCBI to give users a view of the raw data
- document commands to run full ingest pipeline to see curated output
- provide guidance on how to assess data to make specific config changes

tsibley:
> * will need guide for installing `git subrepo` for managing the `ingest/vendored` scripts

How about we vendor `git subrepo` so that our vendored stuff is bootstrapped?

I know the installation docs for `git subrepo` have you clone its repo and source some shell init, but taking a look deeper there's only three files we strictly need (`git-subrepo`, `help-functions.bash`, and `bash+.bash`) and it'd be pretty easy to vendor just those or even pack them into a single file.

(Hell, we could `git subrepo` the `git-subrepo` repo into `ingest/vendored/`, but I worry I might not be welcome anymore if I seriously suggest that…)

(…the other bad suggestion is that `git subtree` doesn't have this issue either…)

Bootstrapping doesn't solve the `git subrepo` usage hurdles, and I know the longer term solution is ditching vendoring entirely in favor of `augur curate` et al., but it would solve the installation hurdle.

victorlin:
> …the other bad suggestion is that git subtree doesn't have this issue either…

I don't think it's a bad suggestion. Wrote more in https://github.com/nextstrain/ingest/issues/3#issuecomment-1920058184.

tsibley:
[...]

PR #195, opened 2024-03-22
Add ingest tutorials

joverlee521:

Tutorial contents are based on the [first draft of the tutorials](https://docs.google.com/document/d/1_16VYT5MU8oXJ4t6HUHp_smx_kgF9OMsORSCldee1_0/edit#heading=h.2q09xocws67v). 

This PR adds two new sections to our tutorials with two separate ingest tutorials 

- [Using a pathogen repo](https://nextstrain--195.org.readthedocs.build/en/195/tutorials/using-a-pathogen-repo/index.html)
    - [Running an ingest workflow](https://nextstrain--195.org.readthedocs.build/en/195/tutorials/using-a-pathogen-repo/running-an-ingest-workflow.html)
- [Creating a pathogen repo](https://nextstrain--195.org.readthedocs.build/en/195/tutorials/creating-a-pathogen-repo/index.html)
    - [Creating an ingest workflow](https://nextstrain--195.org.readthedocs.build/en/195/tutorials/creating-a-pathogen-repo/creating-an-ingest-workflow.html) 

The new sections of tutorials were motivated by the [SAB meeting](https://docs.google.com/document/d/1zDwbn16ZRlMMcKLGWWVJ3lYPNwo5_v_WmgBsHv30_lU/edit?disco=AAABKGP5kEc) that emphasized the need to keep the simple zika-tutorial for new users.

## Related issue(s)

Resolves https://github.com/nextstrain/docs.nextstrain.org/issues/179
Related to https://github.com/nextstrain/docs.nextstrain.org/issues/188

This PR is based on https://github.com/nextstrain/docs.nextstrain.org/pull/191

## Checklist

- [x] Checks pass


j23414 (approved):
Looks reasonable to me

src/snippets/uncurated-ncbi-dataset.rst
```diff
@@ -0,0 +1,37 @@
...
+
+5. Exit the Nextstrain shell to return to your usual shell environment.
+
+.. code-block::
+
+    $ exit
+
+The produced ``ingest/data/raw_metadata.tsv`` will contain all of the fields available from NCBI Datasets.
+Note that the headers in this file use the human readable ``Name`` of the
+`NCBI Datasets' available fields <https://www.ncbi.nlm.nih.gov/datasets/docs/v2/reference-docs/command-line/dataformat/tsv/dataformat_tsv_virus-genome/#fields>`_,
```
jameshadfield, line 36: When designing the ingest workflow did you consider having an intermediate output which was exactly this (`raw_metadata.tsv`, or maybe 2 files if the sequences were split out into a FASTA)? This would remove the need for users to run these 5 steps which in turn would encourage comparing the NCBI data vs the curated data?

The files aren't too large, so I don't think space is a concern, but if it were we could easily mark them with `temp()` and then this section would be "run with `--notemp` to get the raw NCBI data to compare against the curated data".

src/snippets/uncurated-ncbi-dataset.rst
```diff
@@ -0,0 +1,37 @@
+1. Enter an interactive Nextstrain shell to be able to run the NCBI Datasets CLI commands without installing them separately.
```
jameshadfield, line 1: There are many ways to get to the some outcome and docs like these always have to balance covering all the ways to get there vs detailing just one. (Examples are everywhere - we detail nextstrain CLI rather than the snakemake commands it's actually running, we assume non-ambient runtimes.) Sometimes adding little pointers to indicate the links between the different methods would be really helpful. In this example, we are running commands which are largely recreating the steps we just ran in the "running an ingest workflow" section - steps 2+3 are identical to `snakemake --cores 1 --notemp fetch_ncbi_dataset_package`.

I'd change the introduction slightly to indicate this:

```diff
- If you want to see the uncurated NCBI Datasets data to decide what changes
- you would like to make to the workflow, you can run the following:
+ If you want to see the uncurated NCBI Datasets data to decide what changes
+ you would like to make to the workflow, you can download the raw NCBI data
+ by manually running commands very similar to those the pipeline used earlier
+ when "running an ingest workflow"
```

I'd be interested in adopting a consistent approach throughout the docs where we add [little hints](https://docutils.sourceforge.io/docs/ref/rst/directives.html#hint) to explain links between what we're detailing and the other ways it can be / is done. Using this as an example (and based off the text at the end of this snippet), I'm imagining a hint section after the commands such as:

> These commands are actually run by the ingest pipeline with some minor differences. The ingest pipeline restricts the columns to those defined in `config["ncbi_datasets_fields"]`, modifies the header names using `config["curate"]["field_map"]` (which maps the [human-readable strings NCBI uses](https://www.ncbi.nlm.nih.gov/datasets/docs/v2/reference-docs/command-line/dataformat/tsv/dataformat_tsv_virus-genome/#fields) to ones more typical in Nextstrain pipelines), and changes the actual data via a number of curation steps.

I can add similar comments throughout these docs if you are interested in adopting this style, but wanted to discuss here first.

src/tutorials/using-a-pathogen-repo/running-an-ingest-workflow.rst
```diff
@@ -0,0 +1,249 @@
...
+
+Configuring the ingest workflow
+===============================
+
+Now that you've seen the default outputs of the ingest workflow, you can try configuring the ingest workflow to change the outputs.
+If you want to see the uncurated NCBI Datasets data to decide what changes you would like to make to the workflow, you can run the following:
+
+.. include:: ../../snippets/uncurated-ncbi-dataset.rst
+
+We'll walk through an example custom config to include an additional column in the curated output.
```
jameshadfield, line 82: [minor] - can you add a small subheading here to demarcate this from the snippet contents?

src/tutorials/using-a-pathogen-repo/running-an-ingest-workflow.rst
```diff
@@ -0,0 +1,249 @@
...
+
+5. Inspect the ``ingest/results/merged-metadata.tsv`` file to see that it includes the additional columns ``column_A``, ``column_B``, and ``column_C``.
+The records with the ``genbank_accession`` listed in the ``additional-metadata.tsv`` file should have the placeholder data in the new columns,
+while other records should have the default ``?`` value.
+
+Next steps
+==========
+
+* Run the `zika phylogenetic workflow <https://github.com/nextstrain/zika/tree/main/phylogenetic>`_ with new ingested data as input
+  by running
```
jameshadfield, line 242: Note that this won't include any of the data we (may) have just added in "Advanced usage: Customizing the ingest workflow" because the phylo workflow doesn't know about `ingest/results/merged-metadata.tsv`

src/tutorials/using-a-pathogen-repo/running-an-ingest-workflow.rst
```diff
@@ -0,0 +1,249 @@
...
+For example, examining the raw NCBI metadata shows us that ``virus-name`` is a NCBI Datasets field that is not currently downloaded by the default Zika ingest workflow.
+If you wanted this field to be included in your outputs, you could perform the following steps.
+
+1. Create a new build config directory ``ingest/build-configs/tutorial/``
+
+.. code-block::
+
+    $ mkdir ingest/build-configs/tutorial
+
+2. Create a new config file ``ingest/build-configs/tutorial/config.yaml``
```
jameshadfield, line 92: Seeing this big YAML block made me think "whoa, were did all these values come from"? There are a couple of changes you may want to consider here:

* I'd move the text below the YAML to above it. So instead of starting with a big config file¹, start with "the parts of the config we want to change are `ncbi_datasets_fields` (because...), `field_map` (because...) and `metadata_column` (because). 
* Then detail that we copy certain sections from the existing config then add the changes just described.

¹ The config YAML alone takes up the full height of a laptop screen

src/tutorials/creating-a-pathogen-repo/creating-an-ingest-workflow.rst
```diff
@@ -0,0 +1,465 @@
...
+suite of commands to include all of the custom curation steps.
+For now, we've bundled custom scripts into the `ingest repository <https://github.com/nextstrain/ingest>`_ that is then
+vendored in the pathogen-repo-guide using `git-subrepo <https://github.com/ingydotnet/git-subrepo>`_.
+Please do not edit the vendored scripts in ``ingest/vendored`` directly.
+If you run into issues or encounter bugs with the vendored scripts, please `make an issue in the ingest repository <https://github.com/nextstrain/ingest/issues/new/choose>`_.
+Once the bug has been fixed in the original source code, you can follow the `instructions to update the vendored scripts <https://github.com/nextstrain/ingest?tab=readme-ov-file#vendoring>`_.
+
+We highly encourage you to go through the commands and custom scripts used in the ``curate`` rule within ``ingest/rules/curate.smk``
+to gain a deeper understanding of how they work.
+We will give a brief overview of each step and their relevant config parameters defined in ``ingest/defaults/config.yaml`` to help you get started.
```
jameshadfield, line 115: [discussion point, not request for changes]

Here we are mostly taking a template curate pipeline and describing the config values you can change. The other approach would be to detail the curate commands themselves and teach people how to add them to the bash command + abstract their parameters to the YAML config. This reminds me of a similar situation in visualisation where some people build libraries of charts while others expose the methods to build such charts (e.g. see Leland Wilkinson quote [here](https://medium.com/@mbostock/introducing-d3-scale-61980c51545f)).

src/tutorials/creating-a-pathogen-repo/creating-an-ingest-workflow.rst
```diff
@@ -0,0 +1,465 @@
...
+
+We recommend using the `zika ingest workflow <https://github.com/nextstrain/zika/tree/main/ingest>`_ and the
+`mpox ingest workflow <https://github.com/nextstrain/mpox/tree/master/ingest>`_ as example workflows that
+demonstrate our latest developments.
+
+Next steps
+==========
+
+* Learn more about :doc:`augur curate commands <augur:usage/cli/curate/index>`
+* Learn how to create a phylogenetic workflow (TKTK)
```
jameshadfield, line 465: We shouldn't use TKTK in live docs (assuming you aren't going to add the tutorial it in this PR). I'd replace this with "We're planning on writing a similar workflow for the phylogenetic pipeline, but until that's ready you can the best place to learn about these is ..."

src/tutorials/creating-a-pathogen-repo/creating-an-ingest-workflow.rst
```diff
@@ -0,0 +1,465 @@
...
+You can decide whether NCBI Datasets include sufficient data for your pathogen by inspecting the uncurated data from NCBI Datasets CLI.
+
+.. include:: ../../snippets/uncurated-ncbi-dataset.rst
+
+If the data looks sufficient for your pathogen, then use the NCBI Datasets source in your workflow by
+updating the ``ingest/defaults/config.yaml`` config file.
+
+1. Add your pathogen's NCBI taxonomy ID to the ``ncbi_taxon_id`` parameter
+2. If there are other NCBI Datasets fields you would like to include in the download, you can add them to the ``ncbi_datasets_fields`` parameter
+3. Skip to the :ref:`curation-steps`.
```
jameshadfield, line 73: At least in zika we do some extra steps (`format_ncbi_datasets_ndjson`) - do you want to discuss them here? (Maybe I've missed it)

src/snippets/uncurated-ncbi-dataset.rst
joverlee521: No, I had not...This is definitely left over from working with the SARS-CoV-2 data, where I had focused on only using metadata columns that are needed for the workflow to avoid large files.

src/tutorials/using-a-pathogen-repo/running-an-ingest-workflow.rst
joverlee521: Ah, right! I'll update to use `ingest/results/merged-metadata.tsv`.

src/tutorials/creating-a-pathogen-repo/creating-an-ingest-workflow.rst
joverlee521: > The other approach would be to detail the curate commands themselves and teach people how to add them to the bash command + abstract their parameters to the YAML config. 

Hmm, that seems to be stepping into Snakemake tutorial territory...

src/tutorials/creating-a-pathogen-repo/creating-an-ingest-workflow.rst
joverlee521: Yup! Thank you for catching that!

src/tutorials/creating-a-pathogen-repo/creating-an-ingest-workflow.rst
joverlee521: Ah, I had excluded it because it's not a configurable step and seemed like an implementation detail that's not immediately useful to first time users.

src/snippets/uncurated-ncbi-dataset.rst
jameshadfield: For huge datasets this is the right call, but we don't need to build every pipeline to handle the data requirements of SC2.

src/snippets/uncurated-ncbi-dataset.rst
jameshadfield: Yeah, I think there's a range of views out there (including in our group). 

>  I think including multiple/all ways to achieve the task can be distracting and overwhelming for a new user. 

I'm not asking for documenting all of the ways, just adding little hints so that people can join the dots between concepts that may appear separate / independent. In this case it's providing clarifying hints that this section is very related to the section they've just run, even though on the surface they seem completely different.

src/tutorials/creating-a-pathogen-repo/creating-an-ingest-workflow.rst
jameshadfield: I guess I'm wondering how feasible it is for a user to develop an ingest pipeline _without_ knowing Snakemake. (This is the analogue of designing/documenting a charting library.) If that's going to work for a number of users then great! The phrasing you used here ("highly encourage you to go through the [snakemake rules]", "we will give you a brief overview" etc) made it seem to me like knowing Snakemake was a prerequisite. 

Anyway, the docs as they stand are an improvement so I wouldn't want this to hold up their merge. Rather something to think about as we move forward.

joverlee521: > made it seem to me like knowing Snakemake was a prerequisite.

Yup! I have Snakemake listed as a prerequisite for the "Creating an ingest workflow" tutorial at the top:

https://github.com/nextstrain/docs.nextstrain.org/blob/79a0176a4a76e120ebefb42afa90fb073c7d46fc/src/tutorials/creating-a-pathogen-repo/creating-an-ingest-workflow.rst?plain=1#L22-L24

src/snippets/uncurated-ncbi-dataset.rst
joverlee521: I updated the [default config in the pathogen-repo-guide](https://github.com/nextstrain/pathogen-repo-guide/pull/38) to include all of the fields of the raw metadata so that people can remove fields as needed.

src/tutorials/using-a-pathogen-repo/running-an-ingest-workflow.rst
joverlee521: Ah, I remembered that I didn't want to point to the new `merged-metadata.tsv` because I didn't want to go into detail of how to add the new columns to the Auspice config of the phylogenetic workflow. That seems like it needs to be a whole other tutorial...

jameshadfield: Maybe just a line saying "If you've customized the ingest workflow then you may need to modify the phylo workflow to use your ingested data file if it's not `results/metadata.tsv`, and other modifications to your phylo workflow may be needed, as appropriate." Or something

src/snippets/uncurated-ncbi-dataset.rst
joverlee521: Updated tutorial to use the new Snakemake target from the pathogen-repo-guide in https://github.com/nextstrain/docs.nextstrain.org/pull/195/commits/43b0f778647bd1bc1156bc25f964390bcff7e9e9

src/snippets/uncurated-ncbi-dataset.rst
joverlee521: Added your suggested hints in https://github.com/nextstrain/docs.nextstrain.org/pull/195/commits/6019d504aaa114e8c928c27faa3c676fd6a71d18. Feel free to add more!

src/tutorials/using-a-pathogen-repo/running-an-ingest-workflow.rst
joverlee521: Added subheading in https://github.com/nextstrain/docs.nextstrain.org/pull/195/commits/6019d504aaa114e8c928c27faa3c676fd6a71d18

src/tutorials/using-a-pathogen-repo/running-an-ingest-workflow.rst
joverlee521: Updated in https://github.com/nextstrain/docs.nextstrain.org/pull/195/commits/f021e762f2bdab67f081658c7bb044bc77c5887b

src/tutorials/creating-a-pathogen-repo/creating-an-ingest-workflow.rst
joverlee521: Updated in https://github.com/nextstrain/docs.nextstrain.org/pull/195/commits/f021e762f2bdab67f081658c7bb044bc77c5887b

src/tutorials/using-a-pathogen-repo/running-an-ingest-workflow.rst
joverlee521: Updated in https://github.com/nextstrain/docs.nextstrain.org/pull/195/commits/9255b1a21d1564c77fe51be7e31928a2ee78f7f4

joverlee521:
Rebased to include the latest changes merged in #191

src/tutorials/using-a-pathogen-repo/running-an-ingest-workflow.rst
```diff
@@ -0,0 +1,272 @@
...
+    $ cd zika
+
+2. Run the default ingest workflow with the :term:`Nextstrain CLI`.
+
+.. code-block::
+
+    $ nextstrain build ingest
+    Using profile profiles/default and workflow specific profile profiles/default for setting default command line arguments.
+    Building DAG of jobs…
+    [...a lot of output...]
```
tsibley, line 58: All of these new code-blocks which show an example shell session with a prompt and command and optionally output should be `.. code-block:: console` to properly handle the various bits inside with highlighting and the copy/paste button.

joverlee521: That's good to know! I just tried updating to `.. code-block::console` locally, I see a difference in syntax highlighting but the copy/paste behavior is still the same where it only copies the command.

joverlee521: Adding missing code-block languages in https://github.com/nextstrain/docs.nextstrain.org/pull/197

tsibley: > …but the copy/paste behavior is still the same where it only copies the command.

Ah! I thought we were using the [syntax-aware prompt/output exclusion method](https://sphinx-copybutton.readthedocs.io/en/latest/use.html#automatic-exclusion-of-prompts-from-the-copies) of sphinx-copybutton, but [we're using](https://github.com/nextstrain/sphinx-theme/blob/d1ccdba97b571d8655ab48b63fbe9decf9d96071/lib/nextstrain/sphinx/theme/__init__.py#L21-L22) the [pattern-matching exclusion method](https://sphinx-copybutton.readthedocs.io/en/latest/use.html#strip-and-configure-input-prompts-for-code-cells).

All good then.

Issue #201, opened 2024-04-30
Update install docs to indicate supported/required Python version


genehack:
There's a [known issue](https://github.com/nextstrain/augur/issues/1334) that getting Nextstrain (specifically, Augur) install requires a particular Python version, due to dependencies having specific supported versions.

Update the "ambient" option install docs to make this more clear.

victorlin:
Context from [Slack thread](https://bedfordlab.slack.com/archives/C0K3GS3J8/p1714421971153279):

> You can `pip install nextstrain-augur` on any Python [≥3.8](https://github.com/nextstrain/augur/blob/bb86f5629ccddb3bbfca858367cf37a4106e8f49/setup.py#L5), but installing through the Bioconda package won't work on Python ≥3.11 ([a known issue](https://github.com/nextstrain/augur/issues/1334)).

PR #202, opened 2024-04-30
Update ambient install docs to specify Python version

genehack:

Updated the instructions in the ambient section to make it clear that a particular version of Python should be used. I also modified the overall intent of the instructions slightly, so that they direct the user to set up a new Conda environment. The old instructions were oriented more towards adding Nextstrain to an existing Conda enviroment, which (IMO) seems like a less common scenario, and also one more likely to be undertaken by a more experienced user who will be able to adapt these instructions on the fly; doing it this way also side-steps documenting how somebody would need to modify an existing Conda env with an unsupported version of Python.

## Related issue(s)

#201 

## Checklist

- [x] Checks pass

### DIFF
--- a/src/install.rst
+++ b/src/install.rst
@@ -213,7 +213,7 @@ Set up a Nextstrain runtime
 
       The ambient runtime does not require a particular setup method; it will work as long as the programs you wish to use are available.
 
-      The following describes how to accomplish this using a custom Conda environment as an example. You should be familiar with the `basics of Conda <https://conda.io/projects/conda/en/latest/user-guide/getting-started.html>`__ before proceeding.
+      The following describes how to accomplish this by creating a new custom Conda environment, as an example. You should be familiar with the `basics of Conda <https://conda.io/projects/conda/en/latest/user-guide/getting-started.html>`__ before proceeding. If you want to add Nextstrain to an existing Conda environment, please make sure you're using a compatible Python version (we suggest v3.10.13) and activate that environment instead of creating a new one.
 
       .. tabs::
 
--- a/src/install.rst
+++ b/src/install.rst
@@ -213,7 +213,8 @@ Set up a Nextstrain runtime
 
       The ambient runtime does not require a particular setup method; it will work as long as the programs you wish to use are available.
 
-      The following describes how to accomplish this by creating a new custom Conda environment, as an example. You should be familiar with the `basics of Conda <https://conda.io/projects/conda/en/latest/user-guide/getting-started.html>`__ before proceeding. If you want to add Nextstrain to an existing Conda environment, please make sure you're using a compatible Python version (we suggest v3.10.13) and activate that environment instead of creating a new one.
+      .. Suggest Python 3.10.13 because Augur on Bioconda won't resolve dependencies on Python ≥3.11: <https://github.com/nextstrain/augur/issues/1334>
+      The following describes how to accomplish this by creating a new custom Conda environment, as an example. You should be familiar with the `basics of Conda <https://conda.io/projects/conda/en/latest/user-guide/getting-started.html>`__ before proceeding. If you want to add Nextstrain to an existing Conda environment, please make sure you're using Python ≤3.10 and activate that environment instead of creating a new one.
 
       .. tabs::
 
--- a/src/install.rst
+++ b/src/install.rst
@@ -213,7 +213,8 @@ Set up a Nextstrain runtime
 
       The ambient runtime does not require a particular setup method; it will work as long as the programs you wish to use are available.
 
-      .. Suggest Python 3.10.13 because Augur on Bioconda won't resolve dependencies on Python ≥3.11: <https://github.com/nextstrain/augur/issues/1334>
+      .. Suggest Python ≤3.10 because Augur on Bioconda won't resolve dependencies on Python ≥3.11: <https://github.com/nextstrain/augur/issues/1334>
+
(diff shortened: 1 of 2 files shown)


genehack:
[preview](https://nextstrain--202.org.readthedocs.build/en/202/install.html)

src/install.rst
```diff
@@ -213,7 +213,7 @@ Set up a Nextstrain runtime
 
       The ambient runtime does not require a particular setup method; it will work as long as the programs you wish to use are available.
 
-      The following describes how to accomplish this using a custom Conda environment as an example. You should be familiar with the `basics of Conda <https://conda.io/projects/conda/en/latest/user-guide/getting-started.html>`__ before proceeding.
+      The following describes how to accomplish this by creating a new custom Conda environment, as an example. You should be familiar with the `basics of Conda <https://conda.io/projects/conda/en/latest/user-guide/getting-started.html>`__ before proceeding. If you want to add Nextstrain to an existing Conda environment, please make sure you're using a compatible Python version (we suggest v3.10.13) and activate that environment instead of creating a new one.
```
victorlin, line 216: Suggestion: Add a comment linking to why we suggest a specific Python version

```
.. Suggest Python 3.10.13 because Augur on Bioconda won't resolve dependencies on Python ≥3.11: <https://github.com/nextstrain/augur/issues/1334>
```

src/snippets/ambient-setup.rst
```diff
@@ -1,8 +1,11 @@
-1. Activate a Conda environment that you wish to use:
+1. Create a new Conda environment and install Python 3.10.13:
 
    .. code-block:: bash
 
+      conda create -n <your-environment-name>
       conda activate <your-environment-name>
+      conda install --override-channels --strict-channel-priority \
+            -c conda-forge --yes python=3.10.13
```
victorlin, line 8: Is it necessary to split and pin `python` here? I'd assume the `conda install` command in the next step should auto-resolve to a compatible Python version.

I don't have `conda` on macOS so I can't test there immediately, but I checked on Linux via GitHub Codespace and it works fine without this extra step, auto-resolving Python to 3.8.19.

src/install.rst
victorlin, line 216: Suggestion: use the compatibility requirement directly

```diff
- please make sure you're using a compatible Python version (we suggest v3.10.13)
+ please make sure you're using Python ≤3.10
```

src/snippets/ambient-setup.rst
genehack: I originally went down this rabbit hole because my initial install attempt _didn't_ resolve to an installable dep graph because Conda couldn't find a Python version that satisfied everybody. That said, at the point I was using the old version of Conda I had previously had installed, which didn't have the libmamba resolver stuff. 

I repeated the `conda` env build on macOS, without pining the Python version, and it works, and installs 3.8.15. 

Will remove the pining line.

genehack: _Actually..._ you got me to look at this a little bit more, and it turns out `conda create` also supports all the `install` arguments, so I condensed this down a little bit more, as you'll see.

victorlin: This makes it less portable to users who want to install on top of an existing environment. But maybe it's fine to assume those people would know how to pull things out to craft their own `conda install` command. I'm cool with either.

src/install.rst
```diff
@@ -213,7 +213,8 @@ Set up a Nextstrain runtime
 
       The ambient runtime does not require a particular setup method; it will work as long as the programs you wish to use are available.
 
-      The following describes how to accomplish this using a custom Conda environment as an example. You should be familiar with the `basics of Conda <https://conda.io/projects/conda/en/latest/user-guide/getting-started.html>`__ before proceeding.
+      .. Suggest Python 3.10.13 because Augur on Bioconda won't resolve dependencies on Python ≥3.11: <https://github.com/nextstrain/augur/issues/1334>
```
victorlin, line 216: Update the version suggestion here too

```diff
- Suggest Python 3.10.13
+ Suggest Python ≤3.10
```

src/install.rst
```diff
@@ -213,7 +213,8 @@ Set up a Nextstrain runtime
...
+      The following describes how to accomplish this by creating a new custom Conda environment, as an example. You should be familiar with the `basics of Conda <https://conda.io/projects/conda/en/latest/user-guide/getting-started.html>`__ before proceeding. If you want to add Nextstrain to an existing Conda environment, please make sure you're using Python ≤3.10 and activate that environment instead of creating a new one.
```
victorlin, line 217: Suggestion: add a newline between the comment and displayed text. It seems to build fine, but is less readable and my text editor's syntax highlighting thinks the displayed text is part of the comment.

src/snippets/ambient-setup.rst
```diff
@@ -1,19 +1,20 @@
...
+      conda create -n <your-environment-name> \
+            --override-channels --strict-channel-priority \
             -c conda-forge -c bioconda --yes \
             augur auspice nextclade \
             snakemake git epiweeks pangolin pangolearn \
             ncbi-datasets-cli csvtk seqkit tsv-utils
 
+2. Activate the runtime:
+
+   .. code-black:: bash
```
victorlin, line 14: Typo noticed from CI build failure:


```suggestion
   .. code-block:: bash
```

victorlin (changes requested):

src/install.rst
genehack: done.

src/install.rst
genehack: done.

victorlin (approved):

Issue #209, opened 2024-06-17
Read the docs build failures


jameshadfield:
Upstream issue: https://github.com/readthedocs/readthedocs.org/issues/11416

## Progress

Repos affected can be found using GitHub search for [`org:nextstrain  python: "mambaforge-22.9"`](https://github.com/search?q=org%3Anextstrain++python%3A+"mambaforge-22.9"&type=code):

- [x] https://github.com/nextstrain/ncov/pull/1122
- [x] https://github.com/nextstrain/auspice/pull/1798
- [x] https://github.com/nextstrain/nextclade/pull/1500
- [x] https://github.com/nextstrain/docs.nextstrain.org/pull/211

## Solutions/workarounds

1. ⏳ Wait for upstream issue to be resolved
2. ⛔️ Use python instead of mambaforge
3. ⛔️ Pin to `sphinx=7.2`
4. ✅ Bump the mambaforge version to 23.11

---

_Initial commentary_

We recently had multiple failures of the Read-the-docs build in Auspice PRs - https://github.com/nextstrain/auspice/pull/1788, https://github.com/nextstrain/auspice/pull/1779. There were no salient changes to docs related code in those PRs (I believe), and in all cases [our action which builds the docs](https://github.com/nextstrain/.github/blob/master/.github/workflows/docs-ci.yaml) succeeded ([example action log](https://github.com/nextstrain/auspice/actions/runs/9541352737/job/26294445176)). Error (full log [here](https://readthedocs.org/projects/nextstrain-auspice/builds/24712428/)):

```console
critical libmamba Pathname can't be converted from UTF-8 to current locale.
Preparing transaction: ...working... done
Verifying transaction: ...working... failed

CondaVerificationError: The package for sphinx located at /home/docs/.asdf/installs/python/mambaforge-22.9.0-3/pkgs/sphinx-7.3.7-py312h5eee18b_0
appears to be corrupted. The path 'bin/sphinx-apidoc'
specified in the package manifest cannot be found.

CondaVerificationError: The package for sphinx located at /home/docs/.asdf/installs/python/mambaforge-22.9.0-3/pkgs/sphinx-7.3.7-py312h5eee18b_0
appears to be corrupted. The path 'bin/sphinx-autogen'
specified in the package manifest cannot be found.

...
```

The same failure occured on the [latest/stable version of Auspice's RTD](https://readthedocs.org/projects/nextstrain-auspice/builds/24712463/)

[...]

PR #210, opened 2024-06-17
Add `benchmark` to the Snakemake style guide

joverlee521:

Makes it so much easier to identify bottlenecks in workflows if they all just start out with benchmarks!

## Related issue(s)

Prompted by https://github.com/nextstrain/ncov-ingest/issues/446

## Checklist

- [ ] Checks pass


tsibley (approved):

joverlee521:
Read the Docs failure related to #209

corneliusroemer:
I personally dislike the benchmark directive as it adds noise and I never use it, one can just look at logs to see which steps take long - it'd be better if there was a snakemake option to enable benchmarks that didn't add 2 lines to each rule.

victorlin:
+1 on what Jover said, but I get the frustration. It seems like most of the recommendations in our style guide exist to work around less-than-ideal default behavior of Snakemake.

PR #222, opened 2024-08-15
Update filtering and subsampling docs

victorlin:
_[preview](https://nextstrain--222.org.readthedocs.build/en/222/guides/bioinformatics/filtering-and-subsampling.html)_

Prepping for upcoming weighted sampling docs. These changes are applicable to current `augur filter` features.

## Checklist

- [x] Checks pass

### DIFF
--- a/src/guides/bioinformatics/filtering-and-subsampling.rst
+++ b/src/guides/bioinformatics/filtering-and-subsampling.rst
@@ -156,11 +156,56 @@ options. Example:
 Subsampling
 ===========
 
-Another common filtering operation is subsetting of data to achieve a more
-even spatio-temporal distribution or to cut-down data set size to more
-manageable numbers. The filter command allows you to partition the data into
-groups based on column values and sample uniformly. For example, target one
-sequence per month from each country:
+Another common filtering operation is **subsampling**: selection of data using
+rules based on output size rather than individual sequence attributes. These are
+the sampling methods supported by ``augur filter`` and a final section for caveats:
+
+.. contents::
+   :local:
+
+Random sampling
+---------------
+
+The simplest scenario is a reduction of dataset size to more manageable numbers.
+For example, limit the output to 100 sequences:
+
+.. code-block:: bash
+
+   augur filter \
+     --sequences data/sequences.fasta \
+     --metadata data/metadata.tsv \
+     --min-date 2012 \
+     --exclude exclude.txt \
+     --subsample-max-sequences 100 \
+     --output-sequences subsampled_sequences.fasta \
+     --output-metadata subsampled_metadata.tsv
+
+Random sampling is easy to define but can expose sampling bias in some datasets.
+Consider uniform sampling to reduce sampling bias.
+
+Uniform sampling
+----------------
+
+``--group-by`` allows you to partition the data into groups based on column
+values and sample uniformly. For example, sample evenly across countries over
+time:
+
+.. code-block:: bash
+
+   augur filter \
+     --sequences data/sequences.fasta \
+     --metadata data/metadata.tsv \
+     --min-date 2012 \
+     --exclude exclude.txt \
+     --group-by country year month \
+     --subsample-max-sequences 100 \
+     --output-sequences subsampled_sequences.fasta \
+     --output-metadata subsampled_metadata.tsv
+
+An alternative to ``--subsample-max-sequences`` is ``--sequences-per-group``.
+This is useful if you care less about total sample size and more about having
+a fixed number of sequences from each group. For example, target one sequence
+per month from each country:
 
 .. code-block:: bash
 
@@ -173,6 +218,40 @@ sequence per month from each country:
      --sequences-per-group 1 \
      --output-sequences subsampled_sequences.fasta \
      --output-metadata subsampled_metadata.tsv
+
+Probabilistic sampling
+----------------------
+
+It is possible to encounter situations in uniform sampling where the number of
+groups exceeds the target sample size. For example, consider a command with
+groups defined by ``--group-by country year month`` and target sample size
+defined by ``--subsample-max-sequences 100``. If the input contains data from 5
+countries over a span of 24 months, that could result in 120 groups.
+
+The only way to target 100 sequences from 120 groups is to apply **probabilistic
+sampling** which randomly determines a whole number of sequences per group. This
+is noted in the output:
+
+.. code-block:: text
+
+   WARNING: Asked to provide at most 100 sequences, but there are 120 groups.
+   Sampling probabilistically at 0.83 sequences per group, meaning it is
+   possible to have more than the requested maximum of 100 sequences after
+   filtering.
+
+This is automatically enabled. To force the command to exit with an error in
+these situations, use ``--no-probabilistic-sampling``.
+
+Caveats
+-------
+
+For these sampling methods, the number of targeted sequences per group does not
+take into account the actual number of sequences available in the input data.
+For example, consider a dataset with 200 sequences available from 2023 and 100
+sequences available from 2024. ``--group-by year --subsample-max-sequences 300``
+is equivalent to ``--group-by year --sequences-per-group 150``. This will take
+150 sequences from 2023 and all 100 sequences from 2024 for a total of 250
+sequences, which is less than the target of 300.
 
 Subsampling using multiple ``augur filter`` commands
 ====================================================
--- a/src/guides/bioinformatics/filtering-and-subsampling.rst
+++ b/src/guides/bioinformatics/filtering-and-subsampling.rst
@@ -11,8 +11,9 @@ sample data.
 Filtering
 =========
 
-The filter command allows you to select various subsets of your input data for
-different types of analysis. A simple example use of this command would be
+``augur filter`` provides the flexibility to choose different subsets of input
+data for various types of analysis. A simple example would be to select all
+sequences with a collection date in 2012 or later using ``--min-date 2012``:
 
 .. code-block:: bash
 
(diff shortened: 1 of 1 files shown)


src/guides/bioinformatics/filtering-and-subsampling.rst
```diff
@@ -53,9 +53,9 @@ Subsampling within ``augur filter``
 
 Another common filtering operation is subsetting of data to a achieve a more
 even spatio-temporal distribution or to cut-down data set size to more
-manageable numbers. The filter command allows you to select a specific number of
-sequences from specific groups, for example one sequence per month from each
-country:
+manageable numbers. The filter command allows you to partition the data into
+groups based on column values and sample uniformly. For example, target one
+sequence per month from each country:
```
joverlee521, line 58: I just realized that these docs don't explicitly say that the subsampling is applied to the _filtered_ data. It's buried in the `augur filter` help docs, but would be nice to reiterate here.

src/guides/bioinformatics/filtering-and-subsampling.rst
```diff
@@ -86,6 +86,25 @@ total sequences:
...
+   into account:
+
+   1. The actual number of sequences available in the input data. For example,
+      consider a dataset with 200 sequences available from 2023 and 100
+      sequences available from 2024. ``--group-by year --subsample-max-sequences
+      300`` is equivalent to ``--group-by year --sequences-per-group 150``. This
+      will take 150 sequences from 2023 and all 100 sequences from 2024 for a
+      total of 250 sequences, which is less than the target of 300.
+   2. Any sequences force-included by ``--include`` or ``--include-where``. This
+      may result in higher than targeted sequences for some groups.
```
joverlee521, line 106: For me, this brings up the question of how does subsampling interact with `--exclude` or `--exclude-where` options. Can it result in lower than targeted sequences?

joverlee521 (approved):
Nice improvement! I only left non-blocking suggestions, but I think all the additions are very helpful.

src/guides/bioinformatics/filtering-and-subsampling.rst
victorlin: Good point, I'll add more details to this doc.

src/guides/bioinformatics/filtering-and-subsampling.rst
victorlin: [Adding details on order of operations](https://github.com/nextstrain/docs.nextstrain.org/pull/222#discussion_r1720073616) should address this.

src/guides/bioinformatics/filtering-and-subsampling.rst
victorlin: Addressed with 1b7e25f72719eee1e00de29b1c245a459533212d

src/guides/bioinformatics/filtering-and-subsampling.rst
victorlin: Addressed with 1b7e25f72719eee1e00de29b1c245a459533212d

joverlee521 (approved):

src/guides/bioinformatics/filtering-and-subsampling.rst
```diff
@@ -23,20 +24,101 @@ different types of analysis. A simple example use of this command would be
...
      --output-metadata filtered_metadata.tsv
 
-This command will select all sequences with collection date in 2012 or later.
-The filter command has a large number of options that allow flexible filtering
-for many common situations. One such use-case is the exclusion of sequences that
-are known to be outliers (e.g. because of sequencing errors, cell-culture
-adaptation, ...). These can be specified in a separate text file (e.g.
-``exclude.txt``):
+There are several options that allow flexible filtering for many common
+situations. Options can be divided into the following categories:
```
jameshadfield, line 28: [Not blocking]

This reads as 4 exclusive categories, whereas it's two sets of categories (as demonstrated by the 2x2 table). Consider:

Options can be classified based on what data they operate on: {metadata-based, sequence-based}

as well as their behaviour: {standard, force-inclusive}

src/guides/bioinformatics/filtering-and-subsampling.rst
```diff
@@ -23,20 +24,101 @@ different types of analysis. A simple example use of this command would be
...
+
+- **Metadata-based** options work with information available from
+  ``--metadata``.
+- **Sequence-based** options work with information available from
+  ``--sequences`` or ``--sequence-index``.
+- **Standard** options work by selecting or dropping sequences that match
```
jameshadfield, line 34: Would it be fair to divide this set of options into 3 groups: "standard filtering" (preliminary filtering?), "subsampling", and "force-inclusion", and then organise the docs that way? Something like:

**preliminary filtering steps**

**subsampling steps**

**force-inclusion steps**

**alternate approach: subsampling using multiple augur filter commands**

jameshadfield (commented):
Nice additions -- the added "notes" are especially helpful

victorlin:
After working on #223 and finding it hard to fit weighted sampling docs nicely, I've come back here and replaced c334b314ee32fc1682e837faac343b6073aea2e1 with 65f8fe48d57e65d181a66e49840188784140c096. This seems much better.

src/guides/bioinformatics/filtering-and-subsampling.rst
victorlin: Oh that makes sense! I'll try it. Do you think it's still worth distinguishing metadata vs. sequence based filters here? They're already distinguished on the [augur filter docs](https://docs.nextstrain.org/projects/augur/en/latest/usage/cli/filter.html), so I'm thinking no need to add it here.

victorlin: > P.S. I was going to suggest going all out and linking each argument in the table to the corresponding entry in the the augur filter docs, but there's not anchors for each argument.

+1 for https://github.com/nextstrain/augur/issues/1458

src/guides/bioinformatics/filtering-and-subsampling.rst
```diff
@@ -8,35 +8,186 @@ sample data.
...
+       * ``--query``
+     - * ``--min-length``
+       * ``--max-length``
+       * ``--non-nucleotide``
+
+   * - Subsampling
+     - * ``--subsample-max-sequences``
+       * ``--group-by``
+       * ``--sequences-per-group``
+       * ``--group-by``
```
jameshadfield, line 57: Typo - `group-by` is included twice

jameshadfield (approved):
Looks really good! Thanks for humouring my ideas

src/guides/bioinformatics/filtering-and-subsampling.rst
```diff
@@ -8,35 +8,186 @@ sample data.
...
+
+      augur filter \
+        --sequences data/sequences.fasta \
+        --metadata data/metadata.tsv \
+        --min-date 2020 \
+        --include include.txt \
+        --output-sequences filtered_sequences.fasta \
+        --output-metadata filtered_metadata.tsv
+
+  ``Wuhan/Hu-1/2019`` will still be included even if it does not pass the filter
```
genehack, line 152: _non-blocking, nit_

There's no question that it doesn't pass the filter, right? It's from 2019. 

```suggestion
  ``Wuhan/Hu-1/2019`` will still be included even though it does not pass the filter
```

src/guides/bioinformatics/filtering-and-subsampling.rst
```diff
@@ -45,17 +196,15 @@ To drop such strains, you can pass the filename to ``--exclude``:
...
      --exclude exclude.txt \
-     --output-sequences filtered_sequences.fasta \
-     --output-metadata filtered_metadata.tsv
-
-Subsampling within ``augur filter``
-===================================
+     --group-by country year month \
+     --subsample-max-sequences 100 \
+     --output-sequences subsampled_sequences.fasta \
+     --output-metadata subsampled_metadata.tsv
```
genehack, line 203: _non-blocking suggestion_

An additional paragraph here, explaining a bit more about what the output is going to look like in different scenarios, would (IMO) be helpful. For example, what happens if `--group-by country year month` results in more than 100 groups? How does that interact with `--subsample-max-sequences 100`?

genehack: ah, okay, i see the "Caveats" section below addresses this — perhaps a single sentence with a forward pointer (e.g., "See the 'Caveats' section below for some examples of where dataset size and composition can result in getting less than the requested number of outputs.")

genehack (approved):
some non-blocking suggestions.

src/guides/bioinformatics/filtering-and-subsampling.rst
victorlin: Technically `Wuhan/Hu-1/2019` is just a strain name, and the `date` could be anything. We could use more docs on how the date gets inferred from the metadata, but I don't want to give the impression that it can be parsed from the strain name (even if just through a small change in wording).

src/guides/bioinformatics/filtering-and-subsampling.rst
victorlin: Good point. The scenario you've described actually isn't covered by the current caveats section. I realized I forgot to include an entire section on probabilistic sampling which would address this scenario. I'll add it.

victorlin: Added: https://github.com/nextstrain/docs.nextstrain.org/blob/96f68da125c4c34d05b5f5fbf41c22bc4b399179/src/guides/bioinformatics/filtering-and-subsampling.rst#L222-L243

src/guides/bioinformatics/filtering-and-subsampling.rst
victorlin: Fixed: https://github.com/nextstrain/docs.nextstrain.org/blob/96f68da125c4c34d05b5f5fbf41c22bc4b399179/src/guides/bioinformatics/filtering-and-subsampling.rst#L53-L59

src/guides/bioinformatics/filtering-and-subsampling.rst
victorlin: I've [updated](https://github.com/nextstrain/docs.nextstrain.org/compare/96f68da125c4c34d05b5f5fbf41c22bc4b399179..62e01e51d3523b9fff7c633792f83dfaaf1a5efc) the "Preliminary & force-inclusive" section header to "Preliminary & force-inclusive sampling" to fit better with other section headers.

victorlin: After bikeshedding with myself I've determined that "Preliminary & force-inclusive **selection**" contrasts better with **subsampling**. [Updated](https://github.com/nextstrain/docs.nextstrain.org/compare/62e01e51d3523b9fff7c633792f83dfaaf1a5efc..5779a70fc3b8b5cc4c6bcdbf4b01a7360e6dd236). I'm going to stop arguing with myself and merge now.

Issue #216, opened 2024-07-16
Deprecation: Removal of Sphinx context injection at build time


victorlin:
Brief summary from email:

> We are announcing the deprecation of Sphinx context injection at build time for all the projects. The deprecation date is set on Monday, October 7th, 2024. After this date, Read the Docs won't install the `readthedocs-sphinx-ext` extension and won't manipulate the project's `conf.py` file.

## Links

- [Original email ticket](https://support.nextstrain.org/agent/nextstrain/nextstrain/tickets/details/668036000007333001)
- [Blog post on the deprecation](https://about.readthedocs.com/blog/2024/07/addons-by-default/)

## Tasks

- [ ] Read through the blog post
- [ ] Add any action items here

tsibley:
I read thru the "[Read the Docs Addons enabled by default](https://about.readthedocs.com/blog/2024/07/addons-by-default/)" blog post just now. Some notes:

1. We [use](https://github.com/search?q=org%3Anextstrain+%2F%28%3F-i%29READTHEDOCS%2F&type=code) the `READTHEDOCS` template var, so will need to make the build-time config changes noted in the blog post.
2. We'll want to set `sphinx-rtd-theme`'s new (as of 2.1.0rc2) config option `flyout_display = "attached"` in [our Sphinx theme](https://github.com/nextstrain/sphinx-theme) and hide the floating flyout produced by the new addon.
3. We may need to adapt parts of our sphinx-theme to account for changes to the newer sphinx-rtd-theme version

tsibley:
The changes upstream have caused our docs to start rendering with a floating flyout instead of a fixed docs info/version switcher.

victorlin:
I considered keeping the new floating flyout – on wide screens, it makes use of the extra space on the right side of the docs and allows for more sidebar content to be shown. But on narrower screens, it covers our actual docs. Covering sidebar is probably a better trade-off across screen sizes?

tsibley:
The floating flyout drives me batty, as it's continually in the way. Would definitely prefer to keep it in the sidebar.

victorlin:
Restoring the sidebar flyout is not as simple as setting `flyout_display = "attached"` and hiding the RTD addon flyout. I've posted https://github.com/readthedocs/readthedocs.org/issues/11474#issuecomment-2407994264 hoping to find a path forward.

victorlin:
Given that the new flyout addon can only be disabled per project and not per version, these are our options:

[...]

PR #228, opened 2024-10-09
dev: Use conda-forge not the Anaconda default channels

tsibley:
conda-forge is completely open, Anaconda is not.

## Checklist

- [ ] Checks pass

### DIFF
--- a/conda.yml
+++ b/conda.yml
@@ -1,6 +1,6 @@
 name: docs.nextstrain.org
 channels:
-  - defaults
+  - conda-forge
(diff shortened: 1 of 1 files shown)


genehack (approved):

tsibley:
This is failing CI because of a newer version of the RTD theme upon which our own theme is based: the newer version doesn't support the `display_version` config setting. We need to sort out https://github.com/nextstrain/docs.nextstrain.org/issues/216.

PR #251, opened 2025-05-12
Use exec for more robust logging

### LINKED ISSUE #250: Snakemake style guide: Improve section on using `log`
The current example using `2>&1 | tee {log:q}` does not work well in these scenarios which are common throughout our workflows:

1. When there is a piped chain of commands – only the last command's output gets redirected properly.

    Example:

    ```py
    log:
        "logs/extract_from_strain.txt"
    shell:
        r"""
        augur curate passthru \
          --metadata {input.metadata:q} \
        | scripts/extract_from_strain \
        | augur curate passthru \
          --output-metadata {output.metadata:q} \
        2>&1 | tee {log:q}
        """
    ```

2. When a command uses `>` to redirect stdout to an output file – stderr is mixed into stdout and nothing ends up in the console output or log file.

    Example:

    ```py
    log:
        "logs/fetch_general_geolocation_rules.txt"
    shell:
        r"""
        curl {params.geolocation_rules_url} > {output.general_geolocation_rules} \
        2>&1 | tee {log:q}
        """
    ```

    `{output.general_geolocation_rules}` is corrupted:

    ```
      % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                     Dload  Upload   Total   Spent    Left  Speed

      0     0    0     0    0     0      0      0 --:--:-- --:--:-- --:--:--     0Europe/Netherlands/Oisterwijk/	Europe/Netherlands/North Brabant/Oisterwijk
    Europe/Netherlands/Oss/	Europe/Netherlands/North Brabant/Oss
    Europe/Netherlands/Rotterdam/	Europe/Netherlands/South Holland/Rotterdam NL
    ```

## Possible solutions

1. Recommend to either (1) wrap the entire chain of commands in parentheses or (2) move it to a script file. Both of these would allow `2>&1 | tee {log:q}` to capture stderr from all commands. The choice can be context-dependent, but (1) is probably better for cases where many Snakemake variables are used.

    The example above would look like:

    ```py
    log:
        "logs/extract_from_strain.txt"
    shell:
        r"""
        (
            augur curate passthru \
              --metadata {input.metadata:q} \
            | scripts/extract_from_strain \
            | augur curate passthru \
              --output-metadata {output.metadata:q} \
        ) 2>&1 | tee {log:q}
        """
    ```

2. Recommend something like `2> >(tee {log:q} >&2)` to redirect only stderr to the log file.

victorlin:

The previous example using `2>&1 | tee {log:q}` did not work well in these scenarios:

1. When there is a piped chain of commands – only the last command's output gets redirected properly.
2. When a command uses `>` to redirect stdout to an output file – stderr is mixed into stdout and nothing ends up in the console output or log file.

The updated example covers those scenarios.

## Related issue(s)

Closes #250

## Checklist

- [ ] Checks pass

### DIFF
--- a/src/reference/snakemake-style-guide.rst
+++ b/src/reference/snakemake-style-guide.rst
@@ -211,6 +211,7 @@ Log standard out and error output to log files and the terminal
 
 Use `the Snakemake log directive <https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html#log-files>`_ for each rule that writes output to either standard out or error and direct output to the corresponding log file.
 Use the ``tee`` command to ensure that output gets written to both the log file and the terminal, so users can track their workflow progress interactively and use the log file later for debugging.
+Use ``exec`` to capture output from the entire shell block, which is useful if there are multiple commands (e.g. piping).
 
 Example:
 
@@ -220,9 +221,11 @@ Example:
        "logs/filter.txt"
    shell:
        r"""
+       exec &> >(tee {log:q})
+
        augur filter \
            --metadata {input.metadata:q} \
-           --output-metadata {output.metadata:q} 2>&1 | tee {log:q}
+           --output-metadata {output.metadata:q}
        """
 
 Before using ``tee``, ensure that your workflow uses `bash's pipefail option <http://redsymbol.net/articles/unofficial-bash-strict-mode/>`_, so successful ``tee`` execution does not mask errors from earlier commands in the pipe.


jameshadfield (approved):
Looks good. There's at least one other approach we use which achieve this, but good to have a general suggestion:

```
(cat file | a | b | c) 2> {log:q}
```

joverlee521 (approved):
Nice! I had to look up what the `exec ...` was doing and found a nice [breakdown](https://unix.stackexchange.com/a/145654).

victorlin:
Good note on the other approach. I think the `exec` approach is better because (1) it allows the original command to be copy/pasted for running without log and (2) it applies to all commands in the shell block as long as it's the first line (the `()` subshell approach needs to be applied to all commands if there are multiple).

src/reference/snakemake-style-guide.rst
```diff
@@ -220,9 +221,11 @@ Example:
...
        r"""
+       exec &> >(tee {log:q})
+
        augur filter \
            --metadata {input.metadata:q} \
-           --output-metadata {output.metadata:q} 2>&1 | tee {log:q}
+           --output-metadata {output.metadata:q}
        """
 
 Before using ``tee``, ensure that your workflow uses `bash's pipefail option <http://redsymbol.net/articles/unofficial-bash-strict-mode/>`_, so successful ``tee`` execution does not mask errors from earlier commands in the pipe.
```
tsibley, line 231: I believe this paragraph about `pipefail` no longer applies with the updated `exec` recommendation.

PR #252, opened 2025-05-13
Remove paragraph on pipefail with tee

victorlin:

This was meaningful when the recommendation was

    … | tee {log:q}

It no longer makes sense with changes from #251.

There's still an argument to be made that pipefail should be used with pipes in general, but that is more about shell scripting and does not need to be explicitly mentioned given that Snakemake uses bash's strict mode by default.

## Related issue(s)

Follow-up to https://github.com/nextstrain/docs.nextstrain.org/pull/251#discussion_r2087464717

## Checklist

- [x] Checks pass

### DIFF
--- a/src/reference/snakemake-style-guide.rst
+++ b/src/reference/snakemake-style-guide.rst
@@ -228,10 +228,6 @@ Example:
            --output-metadata {output.metadata:q}
        """
 
-Before using ``tee``, ensure that your workflow uses `bash's pipefail option <http://redsymbol.net/articles/unofficial-bash-strict-mode/>`_, so successful ``tee`` execution does not mask errors from earlier commands in the pipe.
-Snakemake uses bash's strict mode by default, so the pipefail option should be enabled by default.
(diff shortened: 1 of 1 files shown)


victorlin:
Pipefail is useful in both cases, but with an explicit recommendation to always pipe with `tee` at the end, I think it made sense to call out that `tee` will mask errors from earlier.

We could still keep a mention on pipefail in the style guide, but it should be a separate entry rather than a note under the logging entry.

genehack (approved):

jameshadfield:
For posterity, and since I hadn't heard of it (although I had used it plenty), bash doesn't have a strict mode but snakemake uses this "[unofficial bash strict mode](http://redsymbol.net/articles/unofficial-bash-strict-mode/)". Unless we override it, of course!

```sh
    #!/bin/bash
    set -euo pipefail
    IFS=$'\n\t'
```

tsibley:
@jameshadfield Snakemake does not set `IFS`. Doing so globally would be very confusing and break a lot of shell programs. Bash's "strict mode" colloquially refers to `set -euo pipefail`. (I know Snakemake's docs link to that "unofficial strict mode" blog post, but as is unsurprising with Snakemake, the docs are not being precise.)

Issue #15, opened 2020-09-25
Fix code block highlighting issues


eharkins:
Some documents from submodules have code blocks with special syntax highlighting for certain types of files / languages. Read the docs doesn't recognize some of them, so if we want to have this formatting work in read the docs / sphinx, we'll need to find a solution for this, whether by removing the highlighting, using an existing solution like a plugin for sphinx, or writing a custom script.

Low priority since it just results in warnings and the rendered code blocks are still readable.

E.g. from current docs build of this repo: 
```
/Users/eharkins/nextstrain/docs.nextstrain.org/src/ncov/docs/customizing-analysis.md: WARNING: Pygments lexer name 'tsv' is not known
/Users/eharkins/nextstrain/docs.nextstrain.org/src/ncov/docs/customizing-visualization.md: WARNING: Could not lex literal_block as "json". Highlighting skipped.
/Users/eharkins/nextstrain/docs.nextstrain.org/src/ncov/docs/customizing-visualization.md: WARNING: Could not lex literal_block as "json". Highlighting skipped.
/Users/eharkins/nextstrain/docs.nextstrain.org/src/ncov/docs/customizing-visualization.md: WARNING: Could not lex literal_block as "json". Highlighting skipped.
```

sachin-suresh-rapyuta:
Any fix for this?

tsibley:
Those ncov docs have undergone a lot of overhauling since this issue was created, including conversion from Markdown to rST.

The JSON examples are now correctly highlighted using `.. code:: json` blocks in rST. But note that the original issue reported here wasn't that `json` wasn't a recognized language, but that lexing it failed (i.e. the JSON was probably invalid).

The TSV examples are simply not marked with a language anymore. This seems fine to me, as it's not clear to me what exactly would be highlighted for a TSV anyway.

Closing as this is no longer an issue.

PR #261, opened 2025-10-03
Update git repo link for bacterial tutorial

kimandrews:
We are [creating a new tuberculosis workflow](https://github.com/nextstrain/tb/pull/17) that will replace the original workflow, but the tutorial "Creating a bacterial phylogenetic workflow" is based on the original workflow. We have already moved the original workflow to a branch called `north-america-archived`, and now we need to update the github link in the tutorial to point to that branch.

## Related issue(s)
[#15](https://github.com/nextstrain/tb/issues/15)

## Checklist

- [ ] Checks pass
- [ ] Update changelog

### DIFF
--- a/src/tutorials/creating-a-bacterial-phylogenetic-workflow.rst
+++ b/src/tutorials/creating-a-bacterial-phylogenetic-workflow.rst
@@ -21,7 +21,7 @@ Setup
 
    .. code-block:: bash
 
-      git clone https://github.com/nextstrain/tb
+      git clone --branch north-america-archived https://github.com/nextstrain/tb
       cd tb
 
    .. note::


jameshadfield:
I think this tutorial is now "archived" (in spirit, not technically). As in, we aren't really suggesting people follow this pattern if they want to analyse bacteria. Or are we? I'd suggest adding a note at the top of the tutorial explaining this situation

kimandrews:
> I think this tutorial is now "archived" (in spirit, not technically). As in, we aren't really suggesting people follow this pattern if they want to analyse bacteria. Or are we? I'd suggest adding a note at the top of the tutorial explaining this situation

@jameshadfield I agree that we should modify the bacterial tutorial. One option would be to change the title and intro of the tutorial to focus on VCF inputs rather than bacteria specifically. See [bad6233](https://github.com/nextstrain/docs.nextstrain.org/pull/261/commits/bad6233c815e8f6b066f151d9c713883aa300d3b) and [preview](https://nextstrain--261.org.readthedocs.build/en/261/tutorials/creating-a-phylogenetic-workflow-with-VCF-input.html) for one possible approach.

PR #272, opened 2026-05-07
Run CI on a schedule

victorlin:

Previously, docs CI was only run on sporadic code changes.

The scheduled runs are meant to catch when any existing links break, at a weekly frequency to reduce noise from potential false positives.

## Related issue(s)

- https://github.com/nextstrain/.github/issues/116

## Checklist

- [x] Checks pass
- [x] ~Update changelog~

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -7,6 +7,11 @@ on:
   pull_request:
   workflow_dispatch:
 
+  # Routinely check that we continue to work in the face of external changes.
+  schedule:
+    # Every Monday at 17:42 UTC / 9:42 Seattle (winter) / 10:42 Seattle (summer)
+    - cron: "42 17 * * 1"
+
 jobs:
   docs:
     name: docs


.github/workflows/ci.yaml
```diff
@@ -4,9 +4,23 @@ on:
...
       - master
+    paths: &paths
+      - .github/workflows/ci.yaml
+      - data/**
+      - src/**
+      - conda.yml
+      - Makefile
+
   pull_request:
+    paths: *paths
```
joverlee521: Listing paths for the docs repo seems like adding overhead for not a lot of gain. I'd expect most files added to this repo to contribute to the docs in some way. I would either remove `paths` or switch to `paths-ignore` to list out specific files to ignore (e.g. .github/dependabot.yml)

victorlin: I started with `paths-ignore` but that ended up being about the same number of entries, to the same point on overhead. I'll just drop all filters.

joverlee521 (approved):

PR #279, opened 2026-07-07
Update download-data options

joverlee521:
_[Preview](https://nextstrain--279.org.readthedocs.build/en/279/guides/share/download-data.html)_

Follow up to all the recent Auspice releases
<https://github.com/nextstrain/auspice/releases/tag/v2.70.0>
<https://github.com/nextstrain/auspice/releases/tag/v2.72.0>

## Checklist

- [x] Checks pass
- [ ] Update changelog

### DIFF
--- a/src/guides/share/download-data.md
+++ b/src/guides/share/download-data.md
@@ -8,9 +8,18 @@ The genbank & GISAID accessions, if applicable, are contained in the metadata do
 Unpublished data is included with permission of the data generators, and does not impact their right to publish.
 Please contact the respective authors (available via the TSV files below) if you intend to carry out further research using their data.
 
+The core nextstrain.org builds should have all download options as long as they
+do not violate the terms of use of the original data source.
+The maintainers of community or Groups builds have the ability to
+[control which assets are available for download](https://docs.nextstrain.org/projects/auspice/en/stable/advanced-functionality/view-settings.html#sharing-control-which-assets-auspice-exposes-for-download)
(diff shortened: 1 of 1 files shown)


src/guides/share/download-data.md
joverlee521, line 1: As a side note, I was surprised to find this in the docs repo instead of the Auspice docs. I'm wondering if this should be moved to Auspice docs...I'm torn because I see how this fits under the overall "Share Analyses" guide, but this is als an Auspice specific feature that should be documented under "Functionality".

jameshadfield: I've always found the split between RTD sub-projects poor from a development point of view, and I can only imagine how confusing it is from a user's point of view. I wish we just had Nextstrain docs.

joverlee521: Hmm, I think the RTD sub-projects is the easiest way for us to have versioned docs per repo.

jameshadfield: Yeah, that's a win. But the organisation structure it imposes is confusing. For instance, [the bioinformatics section](https://docs.nextstrain.org/en/latest/guides/bioinformatics/index.html) is entirely talking about `augur` but it's not part of the versioned augur sub-project. Ditto for your comment which started this thread!

I wish we had a flat structure, perhaps where some parts (e.g. "bioinformatics") were taken from the augur repo and thus versioned. But this is just a rant, I don't think anyone has the bandwidth nor appetite to do any of this at the moment!

